PR #8, opened 2018-12-04
Production checklist

aexvir:


zoo/checklists/views.py
```diff
@@ -0,0 +1,124 @@
...
+    def get_context_data(self, *, object_list=None, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["service_list"] = [
+            {
+                "name": service.name,
+                "owner_slug": service.owner_slug,
+                "name_slug": service.name_slug,
+                "completed_checklist_steps": service.checkmarks.count,
+                "total_checklist_steps": sum(
+                    [len(steps) for tag, steps in STEPS.items() if tag in service.tags]
```
paveldedik: Maybe it would make sense to create a new function/method or property in the `Service` class which would return only the relevant steps. So that you just have something like:

`[len(steps) for tag, steps in service.steps.items()]`

Since you do similar thing here: https://github.com/kiwicom/the-zoo/blob/f312b29c0a5a700d43f1870df648e3fc8bdbcd71/zoo/checklists/views.py#L72-L73

zoo/checklists/views.py
```diff
@@ -0,0 +1,124 @@
...
+
+class UpdateServiceCheklistForm(forms.Form):
+    checked = forms.BooleanField(required=False)
+
+
+def update_service_checklist(request, checklist_item_key, *args, **kwargs):
+    request_data_form = UpdateServiceCheklistForm(request.POST)
+
+    if request.method != "POST":
+        return JsonResponse({"error": ""}, status=403)
```
paveldedik: Perhaps the if statement can be on the first line of this function so that a new form instance is not crated in this case.

And I noticed there is an empty error message, is it meant to be emtpy?

test/checklists/test_steps/test_steps.py
```diff
@@ -0,0 +1,64 @@
...
+    steps = uut.load_all_steps()
+    assert len(steps) == 0
+
+
+def test_steps__with_yaml_files__invalid_file(settings):
+    settings.ZOO_CHECKLISTS_ROOT = Path(f"{test_folder_dir}/steps/invalid_file/")
+
+    with pytest.raises(uut.InvalidStepMetadata):
+        steps = uut.load_all_steps()
+        assert len(steps) == 0
```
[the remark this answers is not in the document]
paveldedik: See also lines 35 and 43.

paveldedik (commented):
Looks good, the code as well as the checklist module itself.

zoo/checklists/views.py
aexvir: Yep, you are right. It's not a huge saving but it makes sense. And I added the error message as well

test/checklists/test_steps/test_steps.py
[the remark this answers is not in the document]
aexvir: Yep, is not necessary. Once the exception is raised the execution doesn't continue. I'll remove it

zoo/checklists/views.py
aexvir: I won't add it to the model as that means adding a dependency that is not really needed for such small thing. But I agree that maybe a function could be good to have, even if is a one-line function.

paveldedik (approved):

paveldedik (approved):

paveldedik (approved):

zoo/checklists/steps.py
```diff
@@ -0,0 +1,56 @@
...
+
+@attr.s(frozen=True, slots=True)
+class Step:
+    tag: str = attr.ib()
+    category_name: str = attr.ib(cmp=False, repr=False)
+    id: str = attr.ib()
+    title: str = attr.ib(cmp=False, repr=False)
+    description: str = attr.ib(cmp=False, repr=False, default=None)
+    help_url: str = attr.ib(cmp=False, repr=False, default=None)
+    is_checked: Optional[bool] = attr.ib(default=None)
```
underyx, line 16: What does None mean in this case? I'm having a hard time imagining an item that's neither complete nor incomplete. Maybe sometimes we have no way to tell? Could you add a comment explaining, if it's worth an explanation?

zoo/checklists/steps.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class IncorrectStepMetadata(Exception):
+    pass
+
+
+def load_all_steps():
+    all_steps = {}
+
+    print(settings.ZOO_CHECKLISTS_ROOT)
```
underyx, line 34: Leftover print 😬

zoo/checklists/steps.py
```diff
@@ -0,0 +1,56 @@
...
+
+    for steps_file_path in settings.ZOO_CHECKLISTS_ROOT.glob("**/*.yml"):
+        try:
+            header, items = yaml.safe_load_all(steps_file_path.read_text())
+        except (ValueError, KeyError, yaml.scanner.ScannerError) as exc:
+            raise InvalidStepMetadata("File format not correct") from exc
+
+        if items:
```
underyx, line 42: I'd flip this to deindent everything below. (if not items: continue)

zoo/checklists/templates/checklists/service_checklist.html
```diff
@@ -0,0 +1,64 @@
...
+  <link rel="stylesheet" href="{% static 'service_checklist.css' %}">
+{% endblock %}
+
+{% block content %}
+{% for category in steps %}
+<h3 class="ui header">{{ category.name }}</h3>
+{% for step in category.steps %}
+<div class="ui fluid {% if step.is_checked %}green{% else %}red{% endif %} segment">
+  <div class="flex-horizontal--space-between">
+    <h4 class="ui marginless header" style="max-width: 640px;">
```
underyx, line 26: Not sure if you intended to leave this inline style here.

zoo/checklists/templates/checklists/service_checklist.html
```diff
@@ -0,0 +1,64 @@
...
+        <div class="sub header">{{ step.description }}</div>
+        {% endif %}
+      </div>
+    </h4>
+    <div class="flex-horizontal">
+      {% if step.help_url %}
+      <a href="{{ step.help_url }}" target="_blank" rel="noopener noreferrer" class="ui tiny basic circular icon button">
+        <i class="help icon"></i>
+      </a>
+      &nbsp;&nbsp;
```
underyx, line 39: Remember what you told me about non-breaking spaces 🤣

Don't need to fix it though 👌

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+        except service_models.Service.DoesNotExist:
+            raise Http404
+
+        context["service"] = service
+        context["steps"] = [
+            {
+                "tag": tag,
+                "name": next(iter(steps.values())).category_name,
+                "steps": [
+                    attr.evolve(
```
underyx, line 69: fuck yeah, immutable code :D

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class ServiceChecklistView(TemplateView):
+    template_name = "checklists/service_checklist.html"
+
+    def get_context_data(self, **kwargs):
+        service_owner_slug = kwargs.get("service_owner_slug")
+        service_name_slug = kwargs.get("service_name_slug")
+
+        if None in [service_owner_slug, service_name_slug]:
```
underyx, line 52: Wat? `if not service_owner_slug or not service_name_slug` is a much more logical way of writing this. Or am I misunderstanding the condition?

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+            raise SuspiciousOperation
```
underyx, line 53: TIL :D

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+            {
+                "tag": tag,
+                "name": next(iter(steps.values())).category_name,
+                "steps": [
+                    attr.evolve(
+                        step,
+                        is_checked=service.checkmarks.filter(step_key=key).exists(),
+                    )
+                    for key, step in steps.items()
+                ],
```
underyx, line 74: I'm not a fan of nested list comprehensions, especially when they go 24 spaces of indentation deep. Extracting this to a utility function could be neat.

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+        return context
+
+
+class UpdateServiceCheklistForm(forms.Form):
+    checked = forms.BooleanField(required=False)
+
+
+def update_service_checklist(request, checklist_item_key, *args, **kwargs):
+    if request.method != "POST":
+        return JsonResponse({"error": "Method not allowed"}, status=403)
```
underyx, line 88: Method not allowed is supposed to be 405, not 403.

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+    owner_slug = kwargs.get("service_owner_slug")
+    name_slug = kwargs.get("service_name_slug")
+
+    if step_checked:
+        try:
+            service = service_models.Service.objects.get(
+                owner_slug=owner_slug, name_slug=name_slug
+            )
+            checkmark = models.Checkmark(service=service, step_key=checklist_item_key)
+            checkmark.full_clean()
```
underyx, line 107: aw 🚿 yiss
✔️

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+            checkmark.save()
+            return JsonResponse({}, status=200)
```
underyx, line 109: Probably would complicate things, but for the record, `HTTP 204 No Content` is nice for these cases; it's a response which marks success with no response body.

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+        models.Checkmark.objects.get(
+            step_key=checklist_item_key,
+            service__owner_slug=owner_slug,
+            service__name_slug=name_slug,
+        ).delete()
+        return JsonResponse({}, status=200)
+
+    except models.Checkmark.DoesNotExist:
+        return JsonResponse(
+            {"error": "The step tried to check does not exist"}, status=400
```
underyx, line 126: HTTP 404?

zoo/checklists/views.py
```diff
@@ -0,0 +1,127 @@
...
+            checkmark.full_clean()
+            checkmark.save()
+            return JsonResponse({}, status=200)
+
+        except service_models.Service.DoesNotExist:
+            return JsonResponse(
+                {"error": "The service specified does not exist"}, status=400
+            )
+
+    try:
```
underyx, line 116: I'd indent all the rest under an `else:` even though it doesn't make a difference for the code execution. It will prevent destructive bugs that can stem from extending the previous part with new branches and forgetting to return.

If we end up extending this code more, it also will make it a more sensible switch once we split the two blocks code to `add_mark()` and `remove_mark()` methods.

zoo/services/forms.py
```diff
@@ -28,6 +29,17 @@ class SentryProjectInput(widgets.TextInput):
     template_name = "services/fields/sentry_input.html"
 
 
+class TagInput(widgets.TextInput):
+    template_name = "services/fields/tag_input.html"
+
+    def get_context(self, name, value, attrs):
+        context = super().get_context(name, value, attrs)
+        context["tags"] = [tag for tag, _ in STEPS.items()]
```
underyx, line 37: same as
```python
context["tags"] = list(STEPS)
```

zoo/services/templates/services/fields/tag_input.html
```diff
@@ -0,0 +1,12 @@
+<div id="tag-input" class="tag-input-container">
+  <tag-input></tag-input>
+</div>
+
+<script>
+  const tagList = [{% for tag in tags %}'{{ tag }}', {% endfor %}]
```
underyx, line 6: Oof, this will break with an XSS the moment we add some tag like `C++ '17` or `InfoSec's project`.

underyx (commented):
Well, I know it's merged already, but I thought I'd leave a couple comments of my own anyway 🙃

There's nothing serious, mostly just informational/educational comments. Ctrl+F `XSS` for the only issue I'm actually a bit worried about.

zoo/checklists/templates/checklists/service_checklist.html
aexvir: Nope, it should be in the less file

zoo/services/templates/services/fields/tag_input.html
aexvir: Unless is explicitly marked as Safe [Django will escape the caracters](https://docs.djangoproject.com/en/2.1/topics/security/#cross-site-scripting-xss-protection). So no XSS should be happening. Unless the sanitization is not correct, but at that point we need to think how much we trust the tool.

aexvir: `C++ '17;alert('test');` gets outputted as `C++ &#39;17;alert(&#39;test&#39;);`

PR #13, opened 2018-12-14
Documentation - add tutorial for custom checks

KrisPavlickova:


docs/index.rst
```diff
@@ -6,8 +6,220 @@
...
+Metadata is a yaml file with a list of errors your checks yield. There are two compulsory parameters in the header:
+
+- :code:`namespace`
+- :code:`category`
+
+Then you list the errors with parameters:
+
+- :code:`id` - a unique identifier (*required*)
+- :code:`title` - a human-readable title (*required*)
+- :code:`description` - additional information about the error (*optional*)
```
aexvir, line 48: Mention that they can also pass information to the description. There is the details object that can be added when yielding the error and in the description it can be rendered.

For example

```yaml
id: something
title: something
description: |
  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.

  {details}
```

docs/index.rst
```diff
@@ -6,8 +6,220 @@
...
+
+The most important parameter of the class is :code:`path`, it holds the information about where the source code is exposed.
+
+All :code:`check_` functions need to return an iterator. The recommended way is to yield generator as soon as you discover errors.
+
+The structure of the generator is: :code:`context.Result("namespace:id", True/False)` where
+
+- :code:`namespace` is the namespace from your metadata file
+- :code:`id` is the id of the raised error from your metadata file
+- :code:`True/False` determines if the error was found (true if it was, false if it was not).
```
aexvir, line 88: Add the details here as well.

docs/index.rst
```diff
@@ -6,8 +6,220 @@
...
+
+Adding custom checks
+====================
+
+If you want to add your own checks to the zoo, you need to create your own project with files:
+
+- metadata (holds the list of errors code can trigger)
+- checks (holds the :code:`check_` functions)
+
+Both of them need to be *accessible to the zoo*.
```
aexvir: Mention that the metadata and checks filenames have to match. We we import the metadata files based on the python check filename.

docs/index.rst
```diff
@@ -6,8 +6,220 @@
 The Zoo
 =======
 
+A smart service catalog providing an overview of your services' development and operations.
+
+Adding custom checks
+====================
+
+If you want to add your own checks to the zoo, you need to create your own project with files:
```
aexvir: Add information about the structure. It should be a python package with two directories inside, `metadata` and `checks`. Inside the `metadata` folder will go the `.yml` files and inside the `checks` folder will go the python functions that actually perform the checks.

docs/index.rst
```diff
@@ -6,8 +6,220 @@
...
+- :code:`True/False` determines if the error was found (true if it was, false if it was not).
+
+Check function structure:
+
+.. literalinclude:: ..\zoo\auditing\standards\dummy_standards\checks\check_constantly.py
+    :language: python
+
+Real life example:
+
+.. code-block:: python
```
aexvir, line 98: I think this example could be simplified. It's too much stuff that it's not directly related to the checks and people might be confused. I know I provided you with the example but I didn't realize that issue. I'll provide you a new one.

aexvir (changes requested):

docs/index.rst
```diff
@@ -6,8 +6,223 @@
 The Zoo
 =======
 
+A smart service catalog providing an overview of your services' development and operations.
+
+Adding custom checks
+====================
+
+If you want to add your own checks to the zoo, you need to create your own project with a python package with two directories inside:
```
aexvir: Yep, here I forgot to add that the package provided must be placed in the directory specified by the `ZOO_AUDITING_ROOT ` environment variable, which is pointing to `zoo/auditing/standards`. Sorry, I just realized this...

aexvir: And do you think it's necessary to explain that `__init__.py` files are needed in the folders? That's implicit in the definition of package, but I'd like to know your opinion on this.

docs/index.rst
```diff
@@ -6,8 +6,223 @@
...
+
+- metadata folder with :code:`yml` files holding the list of errors code can trigger
+- checks folder with python :code:`check_` functions that actually perform the checks
+
+Both of them need to be *accessible to the zoo* and the metadata and checks filenames have to match (the zoo imports the metadata files based on the python check filename).
```
aexvir: I'd change `accessible to` for `importable by`. This can help clarify the package specification.

aexvir (changes requested):

docs/index.rst
KrisPavlickova: I will include it, it is possible that someone not that familiar with python will try to implement it.

aexvir: Makes sense, thanks!

docs/index.rst
KrisPavlickova: Can you please send me the new example 🙏

aexvir: Yep, sorry. I forgot do give you another example. I'd just show something really simple like:

```python
def check_dockerignore(context):
    if not (context.path / "Dockerfile").is_file():
        return

    try:
        dockerignore_text = (context.path / ".dockerignore").read_text()
    except FileNotFoundError:
        dockerignore_text = ""  # no dockerignore is effectively an empty dockerignore

    yield context.Result("dockerignore:create_dockerignore", not dockerignore_text)
    yield context.Result("dockerignore:ignore_dotgit", ".git" not in dockerignore_text)
```

docs/index.rst
```diff
@@ -6,8 +6,98 @@
...
+      title: Use coala to run static analysis checks on your repo
+      description: |
+        Add a `.coafile` so that people can see what rules and checks they should follow
+        when contributing to your repository,
+        and run [coala](https://coala.io) to find and fix these issues automatically.
+
+Checks
+------
+
+The zoo starts all functions in this file that have the prefix :code:`check_` (if you need any helper function just name them anything else).
```
aexvir: I'd change `The zoo starts all functions` with `The zoo detects all functions` as the discovery is decoupled from the execution.

docs/index.rst
```diff
@@ -6,8 +6,98 @@
...
+
+.. literalinclude:: ..\zoo\auditing\standards\dummy_standards\checks\check_constantly.py
+    :language: python
+
+**Real life example:**
+
+.. code-block:: python
+
+    def check_dockerignore(context):
+    if not (context.path / "Dockerfile").is_file():
```
aexvir: The code is not correctly indented. The `def check_dockerignore(context):` should be one level higher than the rest. It was probably copied wrong from GitHub.

aexvir (changes requested):

aexvir (approved):

Issue #19, opened 2019-01-04
Remove issues that no longer apply


aexvir:
Checks can have logic that makes them not run if the it doesn't make sense to run it over the codebase. If there is any issue in the logic that determines if the check should run and the Issue is created there is no way to remove the issue after the logic has been fixed, besides doing it manually.

The runner should detect the checks that are no longer returning results and delete the related issues from the database.

JanBednarik:
It can be solved by deleting Issues with old `last_check` date. There are three reasons why last check date is older than couple of hours - check has been removed, it's broken or does not apply to codebase anymore.

We can create and schedule cleanup task which will handle this. It should do (assuming that we run checks regularly every hour):
- delete all Issues with kind key not found in `KINDS` (removed checks)
- for each kind key: if there is no Issue with last check date newer than 2 hours => check is broken => report to Sentry
- for each kind key: if there is at least one Issue with last check date newer than 2 hours, then delete Issues with last check date older than 2 hours

paveldedik:
Do we need to create a new scheduled task? My idea was to remove no longer relevant issues in the `zoo.repos.tasks.pull` task.

JanBednarik:
`zoo.repos.tasks.pull` is run every hour for every repo and it has different purpose. This cleanup can be run like once a day. And it should be in `zoo.auditing.tasks`, because it's related to auditing Issues.

paveldedik:
Sorry, I should have read your previous comment more carefully. What you are suggesting would probably work and I agree - it makes more sense to include that in a separate cleanup task. However, I'm mainly thinking about situations like:

1. there can be checks that yield results only under some conditions, e.g.
    ```python
    if 'python' in context.languages:
        yield context.Result(...)
    ```
    In such cases, it can happen that python is no longer used in a repository and nonrelevant issues remain in the DB.
2. issues of deleted checks should be removed from the database (basically your first bullet).

I believe both can be handled in the `runner` (which is called from the pull task).

JanBednarik:
Your point 1. is covered by third bullet.

It should not be part of `runner`, because it's not used only on `pull` task, but also in API. Also it would unnecessarily complicate it. It's better to keep this cleanup logic separate if possible.

PR #20, opened 2019-01-09
Add task to cleanup issues periodically

paveldedik:
- remove issues of deleted kind
- remove issues older than configured number of hours (default 2)
- related issue: https://github.com/kiwicom/the-zoo/issues/19

### DIFF
--- a/test/auditing/test_tasks.py
+++ b/test/auditing/test_tasks.py
@@ -1,12 +1,27 @@
 import pytest
+from contextlib import contextmanager
 
+import arrow
 from factory import Faker
 
 from zoo.auditing import tasks as uut
+from zoo.auditing.check_discovery import KINDS
+from zoo.auditing.models import Issue
 from zoo.repos.models import Provider
 
-
 pytestmark = pytest.mark.django_db
+
+
+@contextmanager
+def kinds(*list_of_kinds):
+    for kind in list_of_kinds:
+        KINDS[kind.id] = kind
+    try:
+        yield
+    finally:
+        # remove the kind from global variable KINDS
+        for kind in list_of_kinds:
+            KINDS.pop(kind.id, None)
 
 
 @pytest.mark.parametrize("repository__provider", [item.value for item in Provider])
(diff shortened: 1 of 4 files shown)


JanBednarik:
@aexvir We were discussing this solution as one of options with Pavel. We decided that separate cleanup task will be better.

Some reasons are in issue #19. Maybe most important reason is that solution you suggested will delete also issues for checks which temporarily fails and yields no results (like when check relies on 3rd party API which is temporarily not available). It's better to keep them with old last check date than delete them.

If you want, we can talk about it.

aexvir:
As I see it there is no big deal of having the checks deleted if they are not returning anything. The dev of the check can (and probably should) react if the 3rd party api responds or not and still return something. What could be discussed here is the need of some indication for an error status, which seems like something we'll eventually need.

If a check is not working correctly there is not a big value on having it either way. The responsible of the check working fine and making sure that stuff is returned when it's needed and not returned when is not it's the developer of the check itself.

The periodic task is also not solving this issue either. If nothing is returned the Issue will not be updated and it will be deleted either way, with a delay of two hours, but still delayed. If you really think that delay is important it can be easily added to the solution I propose as well.

On the contrary, having a periodic task deleting all the checks that weren't updated in the last two hours can carry bigger problems, like for some reason there is an error scheduling the pull tasks but having the cleanup tasks being scheduled correctly, in that case we could end up with all our issues removed, and the developer of the checks wouldn't be responsible for this error, as it's not something he can control.

And overall it makes way more sense to do this in the only place where the status of the issues is managed, which is the runner.

JanBednarik:
Let's talk about it on Monday, it will be more effective.

underyx:
> for some reason there is an error scheduling the pull tasks but having the cleanup tasks being scheduled correctly, in that case we could end up with all our issues removed, and the developer of the checks wouldn't be responsible for this error, as it's not something he can control

This is a dealbreaker indeed. I think we should go with Alex's suggestion.

Consider also adding a `gone` status and keeping those around for 7 days or so, and then cleaning them up. The app can then display them in the report: "this issue is gone, will be deleted soon" — but users might get confused by this when the zoo maintainers remove a check.

I scheduled a meeting for Monday 10am where the four of us will design this. @JanBednarik @paveldedik please prepare by thinking of some solution for the problem I quoted.

PR #25, opened 2019-01-29
auditing: cleanup removed or no longer relevant issues

paveldedik:
Introduces a new issue status - `UNKNOWN`, which is the status of issues no longer returned from checks. There are few reasons why this makes sense:

1. it's true, we don't know the status of the issue, it may have been resolved, it may not be relevant anymore, maybe some checks were deleted, we don't know
2. we show these issues in UI in a new "faded" category called `Deprecated Issues` (always the last category) - 
screenshot from 2019-01-29 12-17-47
3. if there are some foreign keys referencing the issue and the issue is only temporarily unavailable, the referenced rows won't be affected
4. the issues are still deleted by a `cleanup_task` if the `UNKNOWN` status doesn't change after X days (currently set to 7 days)
5. we can decide later to let users decide what to do with these issues - a message like "this issue seems to be resolved, is that true?" yes or no option, which can make sense sometimes (e.g. language usage drops below 5 % - then the zoo sets the unknown status but it may still be relevant)

Note that most users will never see this `Deprecated Issues` section in the zoo as this resolves the edge cases discussed in #19


paveldedik:
BTW the CI pipeline failed probably due to some temporary issue. I cannot rerun the pipeline as I don't have access to Kiwi.com on gitlab.com. Can I get access please? @underyx :-)

aexvir:
Discussed IRL, Instead of using an `UNKNOWN` status we should use the standard `deleted` model field so no requirements in the logic are needed. The `deleted` results will just be filtered out everywhere besides the auditing report, where we'll show them faded out until the deletion, just to make the transition easier.

paveldedik:
@aexvir OK, fixed according to what we discussed IRL.

test/auditing/test_tasks.py
```diff
@@ -29,3 +31,39 @@ def test_open_git_issues(mocker, repository, issue_factory):
...
+            kind_key=kind_3.id,
+            deleted=True,
+        ),
+    )
+
+    uut.cleanup_issues()
+
+    assert Issue.objects.filter(kind_key=kind_1.id).first().id == issue_1.id
+    assert Issue.objects.filter(kind_key=kind_2.id).first().id == issue_2.id
+    assert Issue.objects.filter(kind_key=kind_3.id).first() is None
```
aexvir: Just a recommendation, you could use django's `exists` for this purpose
https://docs.djangoproject.com/en/2.1/ref/models/querysets/#exists

test/auditing/test_tasks.py
```diff
@@ -29,3 +31,39 @@ def test_open_git_issues(mocker, repository, issue_factory):
...
+            last_check=weeks_ago,
+            kind_key=kind_3.id,
+            deleted=True,
+        ),
+    )
+
+    uut.cleanup_issues()
+
+    assert Issue.objects.filter(kind_key=kind_1.id).first().id == issue_1.id
+    assert Issue.objects.filter(kind_key=kind_2.id).first().id == issue_2.id
```
aexvir: This should be `get` instead of `filter`. You shouldn't have more than one with that id, right?

aexvir (approved):

Issue #90, opened 2019-04-29
Scan DNS records


aexvir:
Once we have the metadata that tells us where the services are running we should be able to link it with DNS records on our different providers.

aexvir:
Closing this as we already support Route53

Issue #89, opened 2019-04-29
Create new models to hold this information


aexvir:
We'll need to add new models to support this structure, and get rid of the current `Datacenter` which is only a static list of strings currently.

paveldedik:
As far as I understand (I'm not sure I'm using correct terminology everywhere, sorry for that, I hope it's clear anyway):

* each Zoo service can be theoretically deployed to several datacenters (= Amazon, Google, etc.), not sure if we have such services in Zoo, but in my model, I'm assuming it's possible
* Rancher has Stacks which usually represent one Zoo service, but we cannot relay on that, that's just Rancher's way of scoping things, we only care about Rancher services. Not sure, how this works in GCP, but we also probably care only about GCP services
* One Zoo service has usually multiple Rancher/GCP services, we link Zoo service to Rancher/GCP services using docer image url and Gitlab repo url
* Theoretically we even can have a Zoo service with multiple services in Rancher - one handling requests from zoo1.example.com and another one from zoo2.example.com, so there will be two URLs associated with one Zoo service

I think the following model could work (but it's possible I'm missing some important information):

**Datacenter**
- `id`
- `name`
- `region`
- note: we can decide removing this model and use only strings in `ServiceDatacenter`

**ServiceDatacenter**
- `id`
- `service_id` - ID of the Zoo service
- `datacenter_id` - ID of the DataCenter
- note: with this model, it's possible to have one Zoo service associated with multiple datacenters

**ServiceDatacenterMember**
- `id`
- `service_datacenter_id` - ID of ServiceDatacenter
- `name` or/and `email`
- note: we could also add `user_id` field which would be a foreign key to `user` table, not sure it's useful

**ServiceDatacenterService** (wierd name, let's focus on names later when we agree on the structure)
- `id`
- `service_datacenter_id` - ID of ServiceDatacenter
- `name` - name of the service
- `public_url` - URL of the service, can be empty

aexvir:
> each Zoo service can be theoretically deployed to several datacenters (= Amazon, Google, etc.), not sure if we have such services in Zoo, but in my model, I'm assuming it's possible

Yep, that's true and it doesn't even have to be in different providers. There can be multiple datacenters (regions / zones) in the same provider.

> Rancher has Stacks which usually represent one Zoo service, but we cannot relay on that, that's just Rancher's way of scoping things, we only care about Rancher services. Not sure, how this works in GCP, but we also probably care only about GCP services

[...]

PR #104, opened 2019-06-03
Infrastructure mapping (Amazon, Rancher)

### LINKED ISSUE #87: Map rancher services to Zoo services
We need a reliable way for relating Rancher services to Zoo services, we'll gather some metadata from there, like the users that have access to the service (from rancher members), which instances is it running on (we are using EC2s, so check the labels we have there) and which image is it using.

The most reliable link will probably be the docker image itself, which it has a direct relation with a GitLab repository.

paveldedik:
Closes #87, #89, #90

What is not included in this PR:

- GCP
- Cloudflare


zoo/base/settings.py
```diff
@@ -36,6 +37,13 @@
     ZOO_AUDITING_DROP_ISSUES=(int, 7),
     ZOO_SONARQUBE_URL=(str, None),
     ZOO_SONARQUBE_TOKEN=(str, None),
+    AWS_CONFIG=(str, None),
+    AWS_CONFIG_FILE=(str, "/tmp/aws/config"),
+    AWS_SHARED_CREDENTIALS=(str, None),
+    AWS_SHARED_CREDENTIALS_FILE=(str, "/tmp/aws/credentials"),
+    RANCHER_API_URL=(str, None),
+    RANCHER_USERNAME=(str, None),
```
aexvir: Username and password? Shouldn't it be `RANCHER_ACCESS_KEY` and `RANCHER_SECRET_KEY`?

paveldedik: Yes, it should. I'll fix it.

zoo/datacenters/amazon.py
```diff
@@ -0,0 +1,201 @@
...
+    if names:
+        page_iterator = paginator.paginate(LoadBalancerNames=names)
+    else:
+        page_iterator = paginator.paginate()
+
+    for page in page_iterator:
+        yield from page["LoadBalancerDescriptions"]
+
+
+def iter_load_balancers_v2(names=None, profile_name="default"):
```
aexvir: Couldn't this two be merged into one and passing the version as argument? Or it doesn't make sense?

paveldedik: I think it makes sense. I will merge it.

zoo/services/models.py
```diff
@@ -204,23 +200,3 @@ class Meta:
...
-        unique_together = ("provider", "region")
-
-    provider = models.CharField(max_length=100)
-    region = models.CharField(max_length=100, null=True, blank=True)
-
-    def __str__(self):
-        return f"{self.provider} {self.region}" if self.region else self.provider
-
-
-class ServiceQLSchema(DjangoQLSchema):
```
aexvir, line 220: This shouldn't have been deleted, just remove the datacenter related stuff

aexvir (approved):

PR #114, opened 2019-06-12
Make user-agent header in HTTP requests configurable

paveldedik:


aexvir:
I think we could get rid of the third party `retry` dependency as well, as it doesn't really add that much value once we have a single session.

paveldedik:
@aexvir done

aexvir (approved):

PR #118, opened 2019-06-20
Add GCP infrastructure mapping

### LINKED ISSUE #88: Map GCP projects to Zoo services
This is going to be a little bit more complicated than with Rancher I think.

We can start from the same approach, which will be to check the images running on the pods, and if the images are from our registry then try to link it with an existing repository.

Same metadata to be obtained, which right now is people that handles it and instances where is running. It can be checked on the GCP project members and on the cluster information itself.

bayotop:
Fixes #88.


Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM python:3.7-alpine
+# gcloud doesn't support 3.7
+FROM python:3.6-alpine
```
aexvir: This is bad, Google probably won't be adding support for newer versions anytime soon. We discussed it IRL and the best option, both for code and maintainability seems to be to use the APIs directly. There is a [python wrapper](https://github.com/googleapis/google-cloud-python) that already supports this, so we'll have a look at that.

If it's not enough, we'll just use `pyenv` to have multiple python versions, so we can update the main app and still use the `gcloud` client

Dockerfile
```diff
@@ -14,8 +15,16 @@ RUN apk add --no-cache --virtual=.build-deps curl build-base postgresql-dev && \
...
     pip install --no-cache-dir -r requirements.txt -r test-requirements.txt && \
+    # gcloud setup
+    curl https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz > /tmp/google-cloud-sdk.tar.gz && \
+    mkdir -p /usr/local/gcloud && \
+    tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz && \
+    /usr/local/gcloud/google-cloud-sdk/install.sh && \
     apk del .build-deps
 
+ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
+RUN gcloud components install kubectl
```
aexvir: Same for `kubectl`. It might be easier to also use some client library, like [this one](https://github.com/kubernetes-client/python)

paveldedik (approved):

PR #139, opened 2019-08-06
auditing: Make it possible to patch with HTTP requests

paveldedik:


zoo/auditing/runner.py
```diff
@@ -14,13 +14,18 @@
 Result.__new__.__defaults__ = (None,)
 
 Patch = namedtuple("Patch", ["action", "file_path", "content", "previous_path"])
-# set default to `None` for `previous_path`
-Patch.__new__.__defaults__ = (None,)
+# set default to `None` for `previous_path` and `content`
+Patch.__new__.__defaults__ = (None, None)
```
aexvir: This is not needed on python 3.7, you can just change the line 16 to be 

```python
Patch = namedtuple("Patch", ["action", "file_path", "content", "previous_path"], defaults=(None, None))
```

zoo/auditing/runner.py
```diff
@@ -14,13 +14,18 @@
...
+
+RequestPatch = namedtuple("RequestPatch", ["url", "method", "headers", "body"])
+# set default to `None` for `headers` and `body`
+RequestPatch.__new__.__defaults__ = (None, None)
```
aexvir: Same here

zoo/auditing/runner.py
```diff
@@ -14,13 +14,18 @@
...
 
 
 class CheckContext:
     Result = Result
     Patch = Patch
```
aexvir: Would it be a big deal to rename the simple `Patch` to `CodePatch`? Right now is not really clear what's the difference, and I could expect `RequestPatch` to be extending `Patch`

zoo/auditing/utils.py
```diff
@@ -10,5 +18,154 @@ def create_git_issue(issue, user_name, reverse_url):
     issue.save()
 
 
-def get_md5_hash(string):
-    return hashlib.md5(string).hexdigest()
+def generate_random_key(length=512):
+    return "".join(
+        random.SystemRandom().choice(string.ascii_uppercase + string.digits)
```
aexvir: Any reason for not using something like `uuid`?

paveldedik: I've read this post - https://stackoverflow.com/questions/2257441/random-string-generation-with-upper-case-letters-and-digits/23728630#23728630 and thought it would be better

paveldedik: But I didn't pay that much attention to it, there is also apparently a `secrets` module since python3.6

zoo/auditing/runner.py
paveldedik: Renamed, we just need to rename this also in the checks (I will do it) and merge to master at the same time.

zoo/auditing/utils.py
```diff
@@ -10,5 +18,153 @@ def create_git_issue(issue, user_name, reverse_url):
...
+        self.context = context
+
+    def to_dict(self, obj):
+        if isinstance(obj, self.context.CodePatch):
+            return self._patch_to_dict(obj)
+        if isinstance(obj, self.context.RequestPatch):
+            return self._request_patch_to_dict(obj)
+        return obj._asdict()
+
+    def _patch_to_dict(self, patch):
```
aexvir: Rename this to `_code_patch_do_dict`

paveldedik:
```suggestion
    def _code_patch_to_dict(self, patch):
```

aexvir (approved):

PR #142, opened 2019-08-07
Pavel/hide wontfix

paveldedik:
Had to update the discover_checks function which was failing when some patch functions weren't registered in metadata at all. I already fixed this in https://github.com/kiwicom/the-zoo/pull/139 But it should be fixed ASAP.


aexvir (approved):

aexvir:
Thanks @paveldedik

PR #143, opened 2019-08-09
Add a new app libraries to differentiate them from services

paveldedik:


paveldedik:
@aexvir In this pull request, libraries are not yet included on the auditing overview page. That's because we are planning to refactor that page a bit. But If you think I should include the refactoring (at least part of it so that it works), let me know.

aexvir (approved):

PR #145, opened 2019-08-14
Make database options more configurable

paveldedik:


Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM python:3.7-alpine
 
 ENV ZOO_RDS_PRODUCTION_CERT_PATH=/etc/ssl/certs/rds-combined-ca-bundle.pem \
```
aexvir: We can also remove this one now that we are here. It can be set up in the final image

paveldedik: yes, true

paveldedik: The problem is that I would have to parse the CERT PATH from the DATABASE_OPTIONS, so will we download the CERT also in the final image and remove both variables?

Dockerfile
```diff
@@ -1,6 +1,7 @@
...
+    ZOO_DATABASE_OPTIONS=sslmode=verify-full,sslrootpath=/etc/ssl/certs/rds-combined-ca-bundle.pem \
```
aexvir: And this as well, this should go in the final image, so this can be reused with different configs without many changes.

Dockerfile
aexvir: Yep, those variables should be set up on the final deployment. This should work without SSL mode in the DB conn, and we'll add it on the final image.

aexvir: So remove the download of the cert as well

aexvir (approved):

PR #158, opened 2019-09-02
auditing: Refactor audit overview page

paveldedik:
`base` is set to `pavel/add-libraries` - https://github.com/kiwicom/the-zoo/pull/143 right now, it has to be merged first as the refactoring takes into account libraries.


zoo/auditing/assets/js/components/IssueHeader.vue
```diff
@@ -0,0 +1,105 @@
...
+          data-position="bottom right"
+          @click="showPopup"
+        >
+          <i class="material-icons">info_outline</i>
+        </button>
+        <button @click="showProjects" class="ui transparent icon button">
+          <i class="material-icons">{{ menu }}</i>
+        </button>
+        <button @click="selectIssue" class="ui transparent icon button">
+          <i class="material-icons">{{ icon }}</i>
```
aexvir: Please use Semantic UI's ones. Now it shouldn't affect performance as before

zoo/auditing/assets/js/components/IssueList.vue
```diff
@@ -0,0 +1,194 @@
+<template>
+  <div class="issue-list ui segment loading">
```
aexvir, line 2: Check https://semantic-ui.com/modules/accordion.html

paveldedik: I didn't use accordion in the end for the following reasons:

- it expects that the content is rendered just invisible, we would need to render all projects (which is a lot of html)
- it can still be done effectively, just render the content with vue and run `$(element).accordion('refresh')`, which I tried doing
- the problem is that the CSS is very different from `segments` and it looks weird, it would require a lot of changes (e.g. by default, accordion has max width `600 px`)
- I used `ui segments` and nested `ui segment` classes, it looks better and was much easier to do

paveldedik:
@aexvir Fixed issues we discussed IRL. I also updated yarn dependencies.

aexvir (approved):

PR #200, opened 2019-10-10
Coala and black makefile targets

m1n0:


Makefile
```diff
@@ -40,3 +41,9 @@ build-docs:
 
 open-docs:
 	open docs/_build/html/index.html
+
+coala:
+		docker run --rm -it -w /app -v $(curr_dir):/app:cached coala/base:0.11 coala -n
```
paveldedik, line 46: We are using `tox.ini`, where we should run black too. Here, I would just update `test` so that you can pass arguments to tox.

Actually, for black and coala, I just created this issue: https://github.com/kiwicom/the-zoo/issues/201

paveldedik (changes requested):
For now, I would just make the `make test` accept arguments which will be passed to `tox`.

For more complex solution, see https://github.com/kiwicom/the-zoo/issues/201

Makefile
m1n0: I understand, but is having a containerized runner for black and coala an issue? I personally prefer running everything in containers rather than locally, so far I was able to avoid "local" installation with no issues, perhaps there could be both tracks (containerized via makefile targets and local via tox) simultaneously.

aexvir: Agree that running as much stuff as possible inside containers is a good approach, specially with coala which has too many dependencies that are even impossible to install properly on a 2019 os. But also I don't think we should invest much more effort in `coala` and we should drop it as soon as possible.

I will merge this, as it doesn't hurt anything and in fact it will be useful until we remove coala.

paveldedik: We could also move this to tox so that we have it in one place. I also agree that using containers is a good approach, but I don't like having tox coala and make coala running different commands that much.

aexvir: I'd remove `coala` from tox, it's messier to run it there as it's not containerized

paveldedik: Alright, but I still think this is not an ideal solution, as now we need to run `tox`, `make coala` and `make black`/pre commit before pushing.

I don't mind having the same commands twice in tox/make/pre-commit. It's about easiness of use/consistency.

aexvir: We should have everything in `pre-commit` 😆 . I don't see much value on `make black` to be honest, as it's already supposed to be running transparently on `pre-commit`, but I see this just as a temporary help until we get rid of coala and we standardize how we run our tooling.

paveldedik: OK, I agree with that, makes sense. Thanks!

m1n0:
@paveldedik Are you happy with how this is now or do you still want to do the `make test` arguments?

paveldedik (approved):

PR #215, opened 2019-10-21
Better docker image identifier parsing

m1n0:


zoo/analytics/tasks/docker.py
```diff
@@ -1,41 +1,44 @@
...
 def analyze(repository, path):
     """Parse dockerfile for image information about platform."""
-    dockerfile = path / "Dockerfile"
-    if not dockerfile.is_file():
-        return
-    dockerfile.resolve()
-    dfp = DockerfileParser(path=dockerfile.as_posix())
-    for command in dfp.structure:
-        if command["instruction"] != "FROM":
+    for sub_path in DOCKERFILE_PATHS:
```
aexvir: Instead of using `DOCKERFILE_PATHS` I think it could be better to just glob `Dockerfile*`. It's convention to name them like that, and actually with this solution we wouldn't consider `Dockerfile.cypress` for example

zoo/analytics/tasks/utils.py
```diff
@@ -59,3 +61,64 @@ def resolve_unknown_format(version: str) -> str:
...
+
+    TODO: move this to some more appropriate OSS package.
+    """
+
+    # pylint: disable=too-many-instance-attributes # Ignore PyCommentedCodeBear
+
+    def __init__(self, image: str):  # Ignore RadonBear
+        self.full_image_name = image
+
+        self.registry = None
```
maroshmka: maybe just a nitpick, but wouldn't it be better to make this properties? it would help readability - you would read just few lines of code for particular attribute parsing. Also "go to definition" would be easier when you see it in some code as it would point you directly to code parsing it. 

properties could be cached. dunno if guys are using already something, but there are multiple implementation we do not need to do it.

zoo/analytics/tasks/docker.py
```diff
@@ -1,41 +1,44 @@
...
-            yield OS("alpine", for_production=True)
-        elif os_fragment.startswith("alpine"):
-            os_version = parse_version(os_fragment.replace("alpine", "").strip(": "))
-            yield OS("alpine", os_version, for_production=True)
-        elif lang_name in KNOWN_DEBIAN_IMAGES:
-            yield OS("debian", for_production=True)
+            yield DockerImage(
+                name=parsed_image_id.repository_name,
+                version=parsed_image_id.version,
+                for_production=True,
```
maroshmka: What does `for_production` do ? And why is it `True` without determining it somehow ? 

Im not saying we should change it, just curious.

zoo/analytics/tasks/utils.py
```diff
@@ -59,3 +61,64 @@ def resolve_unknown_format(version: str) -> str:
...
+        self.username = None
+        self.repository_name = None
+        self.namespaces = None
+        self.tag = None
+        self.version = None
+        self.os = None
+        self.os_version = None
+        self.alias = None
+
+        if "/" in image:
```
maroshmka: as ai read that parsing, i think the properties should be here. It would be nice to have 1 line comment to what it does.

zoo/analytics/tasks/utils.py
```diff
@@ -59,3 +61,64 @@ def resolve_unknown_format(version: str) -> str:
...
+        else:
+            self.repository_name = image
+
+        if "/" in self.repository_name:
+            self.repository_name, _, self.namespaces = self.repository_name.partition(
+                "/"
+            )
+            self.namespaces = self.namespaces.split("/")
+
+        if self.tag and " as " in self.tag:
```
maroshmka: not sure if `   as   ` is valid (with 2 spaces) in dockerfile. if valid, pls change multiple whitespace to 1 before this check.

zoo/analytics/tasks/utils.py
```diff
@@ -59,3 +61,64 @@ def resolve_unknown_format(version: str) -> str:
...
+
+            match = re.search(r"\d+[\d|\.]*\d", self.os)
+            if match:
+                self.os_version = match.group()
+                self.os = self.os.replace(self.os_version, "")
+        else:
+            self.version = self.tag
+
+    def __str__(self):
+        return str(self.__dict__)
```
maroshmka, line 314: This would actually print all the stuff that object has, right? i think we should print only the attributes we parse here.

zoo/analytics/tasks/docker.py
m1n0: I looked into the codebase and it seems that this flag is used for determining whether a dependency is used in production or not, i.e. when it parses `requirements.txt` for python libs it separates "for production" ones as coming from `requirements.txt` and "not for production" ones coming from other requirement files, e.g. `test-requirements.txt`. So in this case we dont have any clever parsing whether a dockerfile is or is not "for production" so it simply sets it to True every time, as that is typically the case.

Correct me if I am wrong in my assumptions please @aexvir .

zoo/analytics/tasks/utils.py
m1n0: ` as ` is for parsing out aliases, e.g. `python:3.7 as base-image` - I will add comments to the parsing part as proposed above

m1n0: Ah I get you comment now, indeed, multiple whitespaces are valid, I added handling for that case.

zoo/analytics/tasks/utils.py
m1n0: No, `__dict__` on an instance returns attributes only, so works ok in this case.

zoo/analytics/tasks/docker.py
aexvir: Yep, that's correct. We could do some smart categorization based on if it's `Dockerfile` (for prod) and `Dockerfile.whatever` (non prod) but I don't think it's on the scope of the issue you are trying to solve here, and it would raise false positives in projects like our `booking/backend`. So for now this looks good to me 🙂

maroshmka: yes agree, out of scope now. thanks guys

btw i find it a bit misleading. having `for_production=True` for package, because its in `requirements.txt` is weird as the package/service do not need to be in production. i'd maybe just store the name of the file it was parsed from and think about categorazing it later (e.g. - core, test, docs, dev)

zoo/analytics/tasks/docker.py
m1n0: Makes sense, changed

aexvir: But you haven't pushed it yet, right? I can't see the changes 🤔 😅

m1n0: indeed, pushed just now

zoo/analytics/tasks/utils.py
m1n0: Makes sense, although changing it made the parser much longer, its easier to understand and it will be much easier to maintain it.

zoo/analytics/tasks/utils.py
```diff
@@ -58,3 +61,250 @@ def resolve_unknown_format(version: str) -> str:
...
+        self.__registry_host_name = None
+        self.__registry_port = None
+        self.__full_image_name = None
+        self.__full_image_name_no_tag = None
+        self.__username = None
+        self.__full_repository_name = None
+        self.__repository_name = None
+        self.__namespaces = None
+        self.__tag = None
+        self.__version = None
```
aexvir, line 84: I'd probably save this as `latest`, what do you think?

test/analytics/tasks/test_utils.py
```diff
@@ -0,0 +1,52 @@
...
+    assert parsed.namespaces is None
+    assert parsed.tag == "3.7"
+    assert parsed.version == "3.7"
+    assert parsed.full_os is None
+    assert parsed.os is None
+    assert parsed.os_version is None
+    assert parsed.alias is None
+
+
+def test_docker_image_id_parsing_long():
```
aexvir, line 104: Maybe some more test cases? With parts missing and so on

zoo/analytics/tasks/utils.py
aexvir: I think it's better to leave as `None` and particularize for each thing when needed

test/analytics/tasks/test_utils.py
m1n0: Makes sense, I will add some random variations based on real examples

aexvir: Awesome, thank you very much 🙂

aexvir: Glad to hear that :)

zoo/analytics/tasks/utils.py
```diff
@@ -58,3 +61,257 @@ def resolve_unknown_format(version: str) -> str:
         else:
             break
     return ".".join(temp_version)
+
+
+class DockerImageId:
+    """Represents parsed docker image id.
+
+    TODO: move this to some more appropriate OSS package.
```
aexvir:
```suggestion
```

aexvir:
Squashed all the commits into one, as it's showing the progress much more clearly like this.

Also, for future contributions, please try to avoid merging branches and rebasing instead, we use the rebase flow usually, and it's a bit harder to mix both flows.

aexvir:
There are also missing migrations, I'm adding them as well

aexvir (approved):

PR #223, opened 2019-10-25
services: Support multiple environments and URLs

### LINKED ISSUE #47: Allow multiple service URLs
The immediate use case we got reported mentioned staging envs, but there could be some use case where the service would benefit from multiple endpoints, like some common service that is running in different stacks.

### LINKED ISSUE #166: Support multiple environments per Service
Services usually have multiple environments/stages, and the URLs for them aren't the same ones. Some services, like Sentry support easily differentiating between environments within the same UI.

We should add support for that in The Zoo, starting with just some basic, where we would have the possibility to store the same info multiple times per service, and just show some UI differentiating it

Suggested by dido

paveldedik:
- Fix https://github.com/kiwicom/the-zoo/issues/166
- Fix https://github.com/kiwicom/the-zoo/issues/47

TODO?

The thing is that it will introduce breaking changes in the API as we are removing fields from Service model. We could either add 

- [x] Add Environments to HTTP API
- [x] Write tests for it

Or alternatively, write a bunch of getters and setters to the Service model. What do you think @aexvir


paveldedik:
Screenshot from 2019-10-25 19-24-28
Screenshot from 2019-10-25 19-25-52

aexvir:
I think I'd rather modify the API, at least looking towards the future, as we are changing the schema, and it should be possible to query environments independently. So we'd need to define the `ServiceEnvironment` type and add the field to the `Service` type so it can be queried from there.

But maybe that should be another issue or issues, to refine our API, as there are many other discrepancies and missing types.

paveldedik:
Yep, I agree. It shouldn't be that hard, I'll probably do it in this PR. But I agree we have more discrepancies, I'll create issue later for that.

paveldedik:
HTTP API updated, tests passing. @aexvir feel free to review whenever you have time.

zoo/base/assets/js/project_detail.js
```diff
@@ -67,7 +67,7 @@ $(document).ready(() => {
     }
   })
 
-  $('.ui.project-actions.dropdown').dropdown({
+  $('.ui.fade.dropdown').dropdown({
```
aexvir, line 70: The `.ui.project-actions.dropdown` is still valid, and it's more specific 🤔

aexvir: Or am I missing something? 😅

paveldedik: I changed it to `fade` so that I don't have to create a specific `service_detail.js` because of that. I think it doesn't have to be that much specific so that we can use it in components on the page.

I added `fade` class to the `project-actions` dropdown, so it should work. This is just generalization so that I don't have to repeat this code specifically for every rendered dropdown.

zoo/services/migrations/0014_add_environment.py
```diff
@@ -0,0 +1,87 @@
...
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="environments",
+                related_query_name="environment",
+                to="services.Service",
+            ),
+        ),
+        migrations.AlterUniqueTogether(
+            name="environment", unique_together={("service", "name")}
+        ),
+        migrations.RunPython(move_data),
```
aexvir: I'd also add a reverse migration, just in case we have to roll it back 🙂

zoo/base/assets/js/project_detail.js
paveldedik: What I mean is that I would have to repeat:

```js
  $('.ui.XXX.dropdown').dropdown({
    action: 'hide',
    position: 'right',
    transition: 'fade up'
  })
```

Three times as the service detail page contains 3 dropdowns now. And would have to create `service_detail.js` specific for service detail view.

zoo/services/migrations/0014_add_environment.py
paveldedik: Good idea :-)

zoo/base/assets/js/project_detail.js
aexvir: Hmm... got your point. But I don't think any of this has value towards libraries, which share this js file, and we'll end up adding more logic to the `service_detail` page, which is the one that needs all the fancy stuff.

It also makes more sense from the point of keeping a consistent structuring, at least in my opinion, feel free to let me know otherwise.

But I'm ok with leaving it as it is for now and move it more in the future 🙂 I'll leave it up to you

paveldedik: OK, I would leave as it is for now. I think dropdowns make sense in libraries as well as there is already one dropdown (the delete/edit actions), so it should be here even after we create `service_detail.js` specific for Services (and I definitely agree we should at some point - but IMO there is no need for now).

aexvir:
Thanks for adding the reverse migration as well 🙂 
Looks good to me!

aexvir (approved):

PR #224, opened 2019-10-25
Add support for gitlab-ci files in the analytics module

### LINKED ISSUE #188: Add support for gitlab-ci files in the analytics module
Mainly images used on the CI jobs and included templates. Names and versions. It can be useful for making informed decisions about the whole company's CI.

uliana291:
Close: https://github.com/kiwicom/the-zoo/issues/188


test/analytics/test_gitlab_ci.py
```diff
@@ -0,0 +1,55 @@
...
+"""
+
+
+def test_parse_docker_images():
+    expected_images = [("docker", "19.03"), ("kiwicom/black", "19.3b0")]
+    parsed_yaml = yaml.safe_load(YAML)
+
+    images = uut.parse_docker_images(parsed_yaml)
+    for image in images:
+        assert image.name, image.version in expected_images
```
aexvir, line 53: I think that the correct assertion would be that **only** those images are returned, as it might be returning more things.

zoo/analytics/tasks/gitlab_ci.py
```diff
@@ -0,0 +1,43 @@
...
+    for name in parsed_yaml.get("include", {}):
+        yield GITLAB_CI(Path(name).stem)
+
+
+def get_image(image_def):
+    """Get names of docker images from yaml dict."""
+    if image_def is None:
+        return None
+
+    name, version = image_def.split(":")
```
aexvir: What happens when there is an image without tag? like `image: python`. In this case the default tag for `docker` is `latest`, so we should also return that

aexvir (changes requested):
Please check the comments I've left 🙂

PR #233, opened 2019-10-30
fix: Correctly identify the .git folder on the app version

aexvir:
Running the tests for the OSS zoo locally was failing for me, as it wasn't finding the proper folder when being inside a submodule. This should address that issue.

### DIFF
--- a/zoo/utils.py
+++ b/zoo/utils.py
@@ -6,10 +6,9 @@ import raven
 
 def _get_app_version():
     """Return current commit SHA of the app."""
-    file_path = Path(__file__)
-    for directory in file_path.parents:
-        if (directory / ".git").exists():
-            return raven.fetch_git_sha(str(directory))
+    basedir = Path(__file__).parent
+    if (basedir / ".git").exists():
+        return raven.fetch_git_sha(str(basedir))
 
     # if there was no git repo in the parent directories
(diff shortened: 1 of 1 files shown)


zoo/utils.py
```diff
@@ -6,10 +6,9 @@
 
 def _get_app_version():
     """Return current commit SHA of the app."""
-    file_path = Path(__file__)
-    for directory in file_path.parents:
-        if (directory / ".git").exists():
-            return raven.fetch_git_sha(str(directory))
+    basedir = Path.cwd().parent
```
paveldedik: Is this really a good fix? Or what was wrong with the previous ~fix~ code? Because cwd can be completely different from where the `.git` repository is located...

aexvir: Fair enough, I assumed that we'd always run it from the root folder, but you are right, it might be a different case. Restored the `__path__` as root for the `.git` discovery

aexvir: The problem is that `.parents()` would recursively "escalate" up the ancestors, picking up an incorrect folder where to look into the version

paveldedik: Yeah, I thought that is the issue, thank you for the explanation.

paveldedik (approved):

PR #236, opened 2019-11-01
Improve DX - setup docs, fake data, keep local data.

maroshmka:
1. document how to add gitlab auth locally - not sure if it is the right way tho
2. fake data generator + superuser - should I make it more complex? is it worth?
3. keep local postgres data - add volume to not delete data when you rm container

### DIFF
--- a/docs/gitlab_auth_local.rst
+++ b/docs/gitlab_auth_local.rst
@@ -1,0 +1,18 @@
+GITLAB AUTH
+===========
+
+1. Create .env file with `ZOO_GITLAB_URL=https://gitlab.com/`
+2. Go to https://gitlab.com/ -> Settings -> Applications
+3. Create an application with:
+	- read_user scope
+	- redirect uri - http://localhost:20966/accounts/gitlab/login/callback/
+4. Copy client id and secret
+5. Go to http://localhost:20966/admin
+6. Go to `Site` model and set `Domain name` to `localhost`
+7. Go to `Social Applications` and create an entry for Gitlab with client id and secret
+8. Profit
+
+
+
+
+
--- a/zoo/base/management/commands/fake.py
+++ b/zoo/base/management/commands/fake.py
@@ -1,0 +1,40 @@
+import sys
+
+import django
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+
+from ... import settings
+from ....services.models import Service, Status
+
+
+class Command(BaseCommand):
+    help = "Fake data for local dev + creates superuser."
+
+    def handle(self, *args, **options):
+        # todo - is this worth ?
+        if settings.DATABASES["default"]["PASSWORD"] != "postgres":
+            print("You're not on localhost, nothings happening.")
+            sys.exit(1)
+
+        # todo - use faker, create more data ?
+        s = Service(
+            owner="jake",
+            name="my-service",
+            status=Status.PRODUCTION,
+            slack_channel="#plz-hello",
+        )
+        try:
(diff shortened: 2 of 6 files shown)


zoo/base/management/commands/fake.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            s.save()
+        except django.db.utils.IntegrityError:
+            print(f"'{s}' already exists, skipping creation.")
+        else:
+            print("Created 1 service.")
+
+        try:
+            User = get_user_model()
+            User.objects.create_superuser("admin", "admin@myproject.com", "password")
```
aexvir, line 36: I'd avoid domains like `myproject.com`, nothing will probably happen, but I don't like the idea of having specific domains. I usually leave the email field empty when I create the superuser

zoo/base/management/commands/fake.py
```diff
@@ -0,0 +1,40 @@
...
+        except:
+            print(f"Admin, already exists, skipping creation.")
```
aexvir, line 38: No need for f-string here

zoo/base/management/commands/fake.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def handle(self, *args, **options):
+        # todo - is this worth ?
+        if settings.DATABASES["default"]["PASSWORD"] != "postgres":
+            print("You're not on localhost, nothings happening.")
+            sys.exit(1)
+
+        # todo - use faker, create more data ?
+        s = Service(
+            owner="jake",
```
aexvir, line 22: I think that if we use `faker` here it could be more useful, so it generates random data, so running the same command multiple times would add more fake services. But not sure about that

aexvir: Oops, just saw your comment 😄

docs/gitlab_auth_local.rst
```diff
@@ -0,0 +1,18 @@
...
+2. Go to https://gitlab.com/ -> Settings -> Applications
+3. Create an application with:
+	- read_user scope
+	- redirect uri - http://localhost:20966/accounts/gitlab/login/callback/
+4. Copy client id and secret
+5. Go to http://localhost:20966/admin
+6. Go to `Site` model and set `Domain name` to `localhost`
+7. Go to `Social Applications` and create an entry for Gitlab with client id and secret
+8. Profit
+
```
aexvir, line 14: Please remove all this extra empty lines 🙂

zoo/base/management/commands/fake.py
maroshmka: :D I can create 100 services (or it can be option how many). Question is - do you see value in doing so ? Should I add faker to requirements and create more services with proper data? 

There's also some Django integration - https://github.com/joke2k/django-faker - but as I read the "guessers" it is not very smart. I think I would rather manually fake the attributes one by one. But maybe its worth to try 🤷‍♂

aexvir: The main limitation I'm seeing is that there aren't repos to relate the services to... so it's not that useful.

To make it more useful I'd probably run `sync_repos` first, and pick up a random repo from `Repositories.objects.all()`, and in that case I can see more value on using the faker, but as you mentioned, just the `owner` and the `name` should be enough for now.

Of course that would require setting up a `GitHub` or `GitLab` token prior to running this command

maroshmka: oh I get it. I'll try to come up with something :)

aexvir:
Closing this as it was merged via #281

PR #243, opened 2019-11-14
feat: Add support for OpenAPI schemas

### LINKED ISSUE #59: List and describe APIs
Different species have different ways of communicating with each other, so in an environment with many services talking between them it could be beneficial to have the list of APIs associated with services, among some information to help kickstart talking to them, as 

- scope of the API
- specs and or documentation links
- staging / sandbox / production links
- authentication mechanisms
- pagination format
- versioning pattern
- deviations from standards or common practices

aexvir:
Closes #59


Dockerfile
```diff
@@ -1,38 +1,39 @@
...
+RUN yarn production
+
 FROM python:3.7-alpine
 
 ENV DJANGO_SETTINGS_MODULE=zoo.base.settings
 RUN addgroup -S macaque && adduser -H -D -S macaque macaque
 
 WORKDIR /app
+
+COPY --from=fe-builder /app/zoo ./zoo
```
paveldedik, line 20: I don't understand this line. What is in `/app/zoo`? Because in `fe-builder` I don't see anything being copied there... `zoo` is copied to `/app/webpack/source/` if I understand it correctly.

zoo/repos/views.py
```diff
@@ -14,16 +24,61 @@ class RepoList(ListView):
...
+        log.info(
+            "repos.views.openapi.invalid", path=str(path.relative_to(base)), error=err
+        )
+
+
+def openapi_definition(request, provider, repo_id):
+    try:
+        repository = Repository.objects.get(remote_id=repo_id, provider=provider)
+    except Repository.DoesNotExist:
+        return JsonResponse({"error": "repository not found"}, status=400)
```
paveldedik: Doesn't HTTP status `404` make more sense here?

```suggestion
        return JsonResponse({"error": "repository not found"}, status=404)
```

zoo/repos/views.py
```diff
@@ -14,16 +24,61 @@ class RepoList(ListView):
...
+        return JsonResponse(json.loads(redis_conn.get(key)), safe=False)
+
+    specs = []
+
+    with tempfile.TemporaryDirectory() as repo_dir:
+        try:
+            repo_path = download_repository(repository, repo_dir)
+        except (MissingFilesError, RepositoryNotFoundError) as err:
+            log.info("repos.views.openapi.git_error", repo=repository, error=err)
+            return JsonResponse({"error": "error downloading repository"}, status=400)
```
paveldedik: I'm not sure about the status here as well. 400 means it's a client error, but it doesn't seem to be the case here.

zoo/repos/views.py
```diff
@@ -14,16 +24,61 @@ class RepoList(ListView):
...
+
+        for ext in ("json", "yml", "yaml"):
+            for path in repo_path.glob(f"**/*.{ext}"):
+                if any(["k8s" in str(path), "test" in str(path)]):
```
paveldedik: I would make the list of blacklisted strings in paths configurable (doesn't have to be configurable via env variable, but I would definitely put it in a variable at the beginning of the file or probably better in settings.py).

zoo/services/templates/services/service_detail.html
```diff
@@ -188,7 +197,36 @@ <h5 class="ui marginless header">
...
   {% endif %}
+{% endblock %}
 
+{% block extra_components %}
+{% verbatim %}
+<div class="ui stackable grid tab" data-tab="openapi">
+  {% if object.repository %}
+  {% endverbatim %}
+  <div id="openapi-visualizer" data-url="{% url 'repo_openapi' object.repository.provider object.repository.remote_id %}">
+  {% verbatim %}
```
paveldedik: What is intended to be in the `verbatim` block here? It looks like the first verbatim ends on line 206 and the second one starts at line 208.

paveldedik (changes requested):
Nice, this feature is great, check the comments please :-)

Dockerfile
aexvir: The webpack script generates [that directory](https://github.com/kiwicom/the-zoo/blob/master/webpack/webpack.build.js#L9) 🙂

zoo/repos/views.py
aexvir: Yep, agree

zoo/repos/views.py
aexvir: Actually yep, agree

zoo/repos/views.py
aexvir: Also agree

zoo/services/templates/services/service_detail.html
paveldedik: And in fact, it is improperly nested as I can see the following text in the browser:

Screenshot from 2019-11-14 11-30-57

aexvir: It's just adding a `non-verbatim` space inside the `verbatim` block from 203 to 229. Because I need to render the URL. 

The verbatim block is needed because Vue.js has the same templating syntax as Django, and Django would try to resolve it otherwise.

Not a huge fan of the solution, it could be done with fancier approaches, but I don't think it's worth the effort. We'll end up switching to an SPA eventually.

Dockerfile
paveldedik: I see :-)

zoo/services/templates/services/service_detail.html
aexvir: Oops... don't know what I was thinking of with that solution... 😅 it was really bad. Fixed now

paveldedik (approved):

PR #253, opened 2019-11-19
chore: Replace coala with pre-commit hooks

aexvir:


.pre-commit-config.yaml
```diff
@@ -1,7 +1,67 @@
-- repo: local
-  hooks:
-    - id: black
-      name: black-code-formatter
-      language: docker_image
-      entry: --entrypoint black kiwicom/black:18.9b0
-      types: [python]
+---
```
paveldedik, line 1: What about adding `pylint` as well? Not necessary since we have it in tox, but maybe it could be here? Just wanted to mainly raise the question, no need to add it here.

paveldedik: But on second thought, I would probably not add pylint here as in tox it installs dependencies so the checking works there better.

paveldedik (approved):
Awesome, this will be much better than coala and can be easily extended :-)

Feel free to merge if you agree with not including pylint in pre-commit.

.pre-commit-config.yaml
aexvir: We are running `pylint` in CI using the docker image, as we bundle it together. For now it's not an issue.

But thanks for raising the point.

Merging then :)

PR #260, opened 2019-12-06
feat(datacenters): Display mapped infra info on service detail page

### LINKED ISSUE #250: Display mapped infrastructure in the service details page
Right now we are automatically mapping our infrastructure from AWS and GCP, but there is no visible output from it.

The resources should be listed in the service detail UI

paveldedik:
This PR displays only what we currently store in the tables connected to `ServiceDatacenter`, this includes:

- list of members with access to runnning GCP/AWS-rancher project
- list of components (Workflows in GCP/Services in Rancher)

I'm not sure if this is enough for release as the information displayed seem to miss context. It would be nice to add also GCP project information and link to Rancher project or something. Which requires new models etc.

The page currently looks like this:

Screenshot from 2019-12-06 10-09-46

Close https://github.com/kiwicom/the-zoo/issues/250


aexvir:
Wowowo, awesome!
I think the top level categories shouldn't be `AWS` and `GCP` but `Rancher` and `Kubernetes`, as technically we can have GCP nodes in Rancher, and Kubernetes can run on AWS, or Azure, or whatever

aexvir:
The information regarding of if it's in AWS or GCP would be better to have in the instance level, as well as the region, as both k8s clusters and rancher stacks can cover multiple regions/zones

aexvir (changes requested):
Please change the top level categories to `Rancher` and `Kubernetes`

paveldedik:
@aexvir Fixed. I also changed the icon, although I'm not sure about it (it's probably not worth trying to find Rancher/K8s icons)... And also added subheader, not sure I want to keep it, but it's an idea...

Screenshot from 2019-12-06 11-07-23

aexvir:
Hmm, agree that it's not critical to have the `rancher` and `kubernetes` icons, although it's not hard to inline svg's like I did with the Sentry icon. But I'll leave it up to you. Here are the SVGs if you want to do it:

[rancher.svg](https://rancher.com/imgs/brand-guidelines/rancher-logo-cow-blue.svg)
[kubernetes.svg](https://cncf-branding.netlify.com/img/projects/kubernetes/icon/black/kubernetes-icon-black.svg)

If you need to opimize it to make the inlining simpler use https://jakearchibald.github.io/svgomg/

aexvir:
> And also added subheader, not sure I want to keep it, but it's an idea...

It's nice, but unless it's listing all the platforms where is running I'd skip it for now. Like I said on a previous comment, we can have servers from AWS, GCP, Azure, etc in Rancher, associating Rancher with AWS it's specific to our usage

paveldedik:
> It's nice, but unless it's listing all the platforms where is running I'd skip it for now. Like I said on a previous comment, we can have servers from AWS, GCP, Azure, etc in Rancher, associating Rancher with AWS it's specific to our usage

I agree, but I'd also say that it's already very specific to our usage. We are mapping AWS + Rancher infrastructure to our models, in the models, there is no field or any kind of information about Rancher, only AWS. And in the templates, we are saying that these members and components are from rancher if `provider==AWS`.

So I don't know how to solve it, maybe we should add a field to Datacenter model with something like `engine` or whatever, which would be `rancher`.

The model has the following fields right now:

- provider
- region

I'm not sure where I would put this information about rancher. But it feels like it's missing from the models, we just assume rancher in templates.

aexvir:
>  there is no field or any kind of information about Rancher, only AWS

> So I don't know how to solve it, maybe we should add a field to Datacenter model with something like engine or whatever, which would be rancher.

Yep... I see... I definitely think that we'll need that layer as well, call it `platform`, `orchestrator` or whatever, but in order to make it useful for people outside Kiwi it's needed.

But... I don't want to block this PR because of that, so I'm fine with merging it and addressing that issue in a separate PR

paveldedik:
OK, thank you for your insight. I will decide if I do it in this PR or create an issue, but I'll fix the blocking issues first.

paveldedik:
Added k8s and rancher icons. I'll create issue for the things discussed above.

aexvir (approved):

PR #262, opened 2019-12-12
refactor(requirements): Use pip-compile-multi approach

aexvir:


paveldedik (approved):
Pretty good solution for conflicting package versions :-)

Issue #254, opened 2019-11-26
Add support to read .zoo.yml while scanning a git repo


imunitic:
Make zoo scan for .zoo.yml file in a git repo. This file should contain all necessary metadata about the project being scanned. Zoo upon reading this file can create or update a project in its registry.

If a project already exists in the registry zoos behavior should be configurable to either overwrite or do nothing. Something like override-existing: false in the yaml file.

The exact content and structure of .zoo.yml is still up for debate.

imunitic:
Here is a first draft of how .zoo.yml could look like.
```yaml
zoo:
  register:
    type: service
    name: hello-world-service
    owner: johndoe
    impact: profit
    status: production
    repository: https://git/hello-world-service.git
    documentation-url: https://docu.com/hello-world-service
    slack-channel: http://slack.com/channel
    sentry-project: http://sentry.com/hello-world-service
    sonarqube-key: hello-world-service
    pager-duty-url: https://pager.com/pager/hello-world-service
    tags:
      - tag1
      - tag2
      - tag3
    environments:
      - staging:
          dashboard-url: https://staging.dashboard-url.com
          service-url: https://staging.service-url.com
      - production:
          dashboard-url: https://production.dashboard-url.com
          service-url: https://production.service-url.com
  actions:
    on-exists: update # possible values update | skip 
```

For now the file has two sections called register and actions. Register is where the metadata for the project is located and actions is used to instruct zoo how to behave on specific events. 

For now there is only one event called on-exists. This instructs zoo what to do if the project is already registered. Possible values are update and skip.

aexvir:
Hmm, I like the idea of being able to store this info in the repos. It would help with a more precise discovery and the info might be easier to be kept up to date.

Just a couple thoughts:

* Not so sure about the actions part, I'd rather have a fixed logic for it, if there is a yaml file we just respect it, populating our database with it. I don't think the complexity of having to maintain the actions part it's worth.
* In that case I'd remove the `register` level as well
* I'd remove the `zoo` level completely, it just adds an extra level without any big advantage
* The environments can support multiple URLs, so it will have to be an array:

```yaml
environments:
    - staging:
       dashboard-url: https://staging.dashboard-url.com
       urls:
           -https://staging.service-url.com
```

imunitic:
With the changes you proposed the yaml file would looke like the following.

```yaml
type: service
name: hello-world-service
owner: johndoe
impact: profit
status: production
repository: https://git/hello-world-service.git
[...]

PR #269, opened 2020-01-13
 Support for .zoo.yml file

imunitic:
Added support for the feature discussed here: https://github.com/kiwicom/the-zoo/issues/254


test/repos/tasks/test_sync_zoo_file.py
```diff
@@ -0,0 +1,160 @@
...
+                "health_check_url": "http://production.test_proj1/health_check",
+            },
+        ],
+    }
+
+    proj = {"id": 11, "provider": "github"}
+
+    uut.update_or_create_service(data, proj)
+    try:
+        service = Service.objects.filter(owner=data["owner"], name=data["name"]).first()
```
paveldedik: `first()` will never trigger an error, you'd have to use `get(owner=..., name=...)` I think.

zoo/repos/tasks.py
```diff
@@ -82,3 +87,75 @@ def pull(reference, provider):
...
+    for env in data["environments"]:
+        e = Environment(service=service, name=env["name"])
+        e.dashboard_url = env["dashboard_url"]
+        e.service_urls = env["service_urls"]
+        e.health_check_url = env["health_check_url"]
+        e.save()
+
+
+def get_zoo_file_content(proj: Dict) -> Union[str, None]:
+    if proj["provider"] == "gitlab":
```
paveldedik: You should probably use `zoo.repos.utils.get_scm_module` here:

```python
provider = get_scm_module(proj["provider"])
provider.get_file_content(...)
```

paveldedik (changes requested):
Good job, I like the usage of `jsonchema`, overall looks pretty solid.

Aside from my comments, two things:

1. It would be nice to have this `.zoo.yml` documented somewhere in README
2. squash the commits into one - we try to do `one atomic change - feature/bugfix` corresponds to `one commit`, also one of the commits contains crazy body - this will be used when creating changelog automatically and it would most likely look really ugly. I would just add reference to the issue and/or add the detailed description to this pull request. I don't mind huge descriptions in commit messages, but the yaml example should not be there, that's specification which should be in README or something.

test/repos/tasks/test_sync_zoo_file.py
```diff
@@ -0,0 +1,160 @@
...
+    except Service.DoesNotExist:
+        pytest.fail("Service object not found!!! This should not happen!!!")
```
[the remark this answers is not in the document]
imunitic: Yeah I agree on this. I will just let it fail.

zoo/repos/github.py
```diff
@@ -79,6 +81,14 @@ def get_languages(remote_id):
     return langs_percent
 
 
+def get_file_content(github_id, path, ref="master"):
+    try:
+        proj = get_project(github_id)
+        return decode(proj.get_contents(path, ref).raw_data["content"])
+    except Exception as e:
+        raise FileNotFoundError(e)
```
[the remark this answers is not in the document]
imunitic: No it is not a good idea to have a catch all clause. I just couldn't find documentation on what kind of exception the library throws. I will look at the source code and change it accordingly.

zoo/repos/tasks.py
imunitic: Good catch. I'll change it.

test/repos/tasks/test_sync_zoo_file.py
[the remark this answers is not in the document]
imunitic: I will just check if the service is None then :)

zoo/repos/github.py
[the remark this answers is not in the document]
imunitic: So the fix here is to just catch RepositoryNotFoundError error and let it fail on all other exceptions.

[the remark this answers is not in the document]
paveldedik: BTW you can also use `raise SomeError from exc`
see https://stackoverflow.com/questions/24752395/python-raise-from-usage

README.md
```diff
@@ -59,6 +59,62 @@ Access web locally:
 Check `Makefile` for shell commands if you want to run them with modified
 parameters.
 
+### .zoo.yml file support for repositories
```
paveldedik: I would create a new sub-header `Configuration of repositories` and I would add this there. IMO it shouldn't be under `Development`. Or feel free to suggest something else.

README.md
```diff
@@ -59,6 +59,62 @@ Access web locally:
...
+
+.zoo.yml is a file that can be placed in a repository which zoo will if found  while scanning repositories
```
paveldedik: This sentence is hard to understand and there is a missing word. I would probably change it to something like

> Repositories scanned by the Zoo may contain the `.zoo.yml` file. This file contains additional information about the scanned repository and how and where it's used in production. If the Zoo finds this file in the root of the project, it will read it and create a `Service` or a `Library` within the Zoo with the provided information based on the collected data automatically. Otherwise the data have to be added to the Zoo manually.
>
> Here is an example of this file:

And I would put just an example here, not the schema itself which you would have to keep synced with the one you have in the source code. The example doesn't have to include all fields, you can add a link to full example/schema from here. The important thing is that developers won't understand the fields, I myself don't know what `sentry_project` should be... should it be full link or just the part after base URL? And what about `slack_channel`, should it be with `#` or without? You get the point. This has to be documented.

paveldedik (changes requested):

README.md
imunitic: Yes you are right here. I'm really bad at writing documentation :)

README.md
imunitic: done! :)

README.md
imunitic: Your explanation was perfect so i just copied it if you dont mind :)

paveldedik (changes requested):
Looks good, just two more things and we can merge it :-)

paveldedik (approved):
Looks good to me! Thank you :-)

PR #272, opened 2020-01-20
feat(services): Increase the slack channel name limit to 80 chrs

aexvir:
Slack's channel name limit is no longer 22 chrs
https://slackhq.com/new-slack-features-search-calls-channels


paveldedik (changes requested):
:+1: You need to generate migrations as well.

paveldedik (approved):

PR #276, opened 2020-01-30
Remove PyICU and fix paths to requirements in setup.py

paveldedik:


requirements/base.txt
```diff
@@ -1,13 +1,11 @@
-# SHA1:728467968860943ba7eecd70ca33a9034bbc00ad
 #
-# This file is autogenerated by pip-compile-multi
```
aexvir, line 3: You didn't use `pip-compile-multi`?

paveldedik: Didn't, just about to fix that, forgot we use it now.

setup.py
```diff
@@ -2,10 +2,10 @@
 
 from setuptools import find_packages, setup
 
-with open("base.txt") as f:
```
aexvir, line 5: Hmm... we don't keep the `requirements` folder when copying to the Docker image, that's why it was changed here

paveldedik: In Dockerfile, we use `pip install -r ...`, but e.g. `tox` is installing via `setup.py` and it was broken. This fixes it, or is it a bad fix?

aexvir: We also do `pip install -e .`, but from what I can see in the CI it works fine. So let's merge it

aexvir (approved):

PR #277, opened 2020-01-31
fix(auditing): Allow multiple checks use the same patch

paveldedik:
If two or more checks are linked to the same patch (via metadata), than only patch for one of the checks will work, the rest of them will return `TypeError: 'NoneType' object is not callable` when the `Kind.apply_patch` method is called. That's because `kind.apply_patch` will be `None` because the kind key will be missing from the `PATCH` variable.

It makes sense that you can have two checks which can be fixed by the same patch.


aexvir (approved):
Good catch 🙂

PR #279, opened 2020-02-03
fix(services): KeyError when sorting problemanic Sentry issues

paveldedik:
There are SentryIssues with category `None` (the model allows it), this fixes KeyErrors.


aexvir (approved):
Approving this, as it doesn't make sense to get issue categories we don't know how to sort, but maybe we should disallow Null values?

paveldedik:
I'm not sure, maybe it's not a problem to store them in the DB anyway. We are not showing them with this filtering anyway.

PR #281, opened 2020-02-05
Better dev data

blfpd:
Screen Shot 2020-02-05 at 15 23 40

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ The Zoo also provides analytics about how dependency usage and its versions evol
 ### Initial setup
 
 - Create a database: `$ make migrate`
-- Create a superuser: `$ make superuser`
+- Create a superuser and fake data: `$ make fake`
 - Create `.env` file as documented below
 
 ### How to run
@@ -42,6 +42,7 @@ The Zoo also provides analytics about how dependency usage and its versions evol
 - Stop and/or delete data: `$ make destroy`
 - Django shell: `$ make shell`
 - Containers logs: `$ docker-compose logs`
+- Generate fake data + superuser `$ make fake`
 
 Access web locally:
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -32,7 +32,7 @@ The Zoo also provides analytics about how dependency usage and its versions evol
 ### Initial setup
 
 - Create a database: `$ make migrate`
-- Create a superuser: `$ make superuser`
+- Create a superuser and fake data: `$ make fake`
```
aexvir, line 35: So, from what I see `make superuser` will still work, it's just that if not used we'll create one with `admin:password` credentials, right?

aexvir:
As I mentioned in the #236 PR, what do you think about:
> To make it more useful I'd probably run sync_repos first, and pick up a random repo from Repositories.objects.all()

when creating services?

README.md
blfpd: That’s the idea.

aexvir: Great

aexvir:
> As I mentioned in the #236 PR, what do you think about:
> 
> > To make it more useful I'd probably run sync_repos first, and pick up a random repo from Repositories.objects.all()
> 
> when creating services?

Alright, so this shouldn't be a reason to block this PR, we'll discuss possible approaches for filling meaningful repo data

aexvir (approved):

zoo/factories.py
```diff
@@ -42,9 +50,14 @@ class Meta:
...
 
     owner = Faker("user_name")
-    name = Faker("domain_word")
-    impact = choice(["profit", "customers", "customers"])
-    repository = SubFactory(RepositoryFactory)
+    name = Faker("slug")
+    status = LazyFunction(lambda: choice([item.value for item in Status] + [None]))
+    impact = LazyFunction(lambda: choice([item.value for item in Impact] + [None]))
+    repository = SubFactory(
+        RepositoryFactory, owner=SelfAttribute("..owner"), name=SelfAttribute("..name")
```
blfpd, line 57: For now we’ll just create new repositories according to owner name and service name.

PR #282, opened 2020-02-06
Create Resources app and view

### LINKED ISSUE #275: Get rid of analytics and libraries pages
They aren't that useful in the current state, as it barely allows proper filtering nor discovery.

We should have a resources page, where we'll list all our currently analyzed resources, so they are easier to discover, and also add to the service view a tab for listing the resources used by that service in particular.

I'm talking about something like this:

New Wireframe 1

With a detailed view similar to this:

New Wireframe 2

blfpd:
Preview of the Resources page

Closes https://github.com/kiwicom/the-zoo/issues/275


dlukeomalley:
@batisteo I looked into the `Pull request / Bento (pull_request)` failure and it was caused by a bug in Bento. We've cut a new release (0.10.1) and updated the Action. **Things should work when you push your next commit and the Action runs again**. Hope this didn't block you, please do let me know if the issue persists or I can do anything to help.

CC @underyx @nbrahms

aexvir:
@batisteo why did the favicon files got deleted?

zoo/base/templatetags/utils.py
```diff
@@ -53,6 +54,11 @@
         r"E": "orange",
         r"F": "red",
     },
+    "language": {
+        DependencyType.JS_LIB.value: "yellow",
+        DependencyType.PY_LIB.value: "green",
+        "Go Library": "teal",
```
paveldedik: Shouldn't you create contant variable for `Go Library` too?

zoo/base/templatetags/utils.py
```diff
@@ -105,7 +111,9 @@ def settings_value(name):
 
 
 @register.filter
-def percent(value):
+def percent(value, denominator=None):
+    if denominator:
+        value = value / denominator
```
paveldedik:
```suggestion
        value /= denominator
```

zoo/resources/templates/dependency_detail.html
```diff
@@ -0,0 +1,92 @@
...
+  <a class="section" href="{% url 'library_overview' %}">Libraries</a>
+  <i class="right angle icon divider"></i>
+  <div class="active section">{{ object.name }}</div>
+</div>
+
+<header>
+  <div class="labels" style="float: right">
+    <div class="ui {% label_color 'language' object.type %} large xhorizontal label">
+      {{ object.type|cut:" Library" }} </div>
+
```
paveldedik:
```suggestion
      {{ object.type|cut:" Library" }}
    </div>
```

zoo/resources/templates/dependency_overview.html
```diff
@@ -0,0 +1,106 @@
...
+{% url 'language_overview' as language_overview_url %}
+
+<nav class="ui container secondary pointing menu">
+  <a
+    href="{{ library_overview_url }}"
+    class="{% if request.path == library_overview_url or request.path == dependency_overview_url %}active {% endif %}item">
+    <i class="book icon"></i>
+    Libraries
+  </a>
+  {% comment %} <a href="{% url 'docker_template_overview' %}"
```
paveldedik: This is probably WIP, just reminder to remove the comments before the final version.

paveldedik (changes requested):
Good job, looks good. Please check my comments and also, in the end, create only two commits, one adding the main feature, second one adding the `if not DEBUG: ...` thing to fix the error when running tests for views.

paveldedik:
I will also want to see it deployed in sandbox, but we will do that after you fix the issues me and alex wrote. I can help with deploying to sandbox.

tox.ini
```diff
@@ -19,7 +19,7 @@ setenv =
 deps =
   -rrequirements/base.txt
   -rrequirements/test.txt
-commands = pytest {posargs:} --cov zoo test
+commands = pytest {posargs:} --cov zoo test --cov-report=term:skip-covered
```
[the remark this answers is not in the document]
blfpd: Sorry, it’s not related. I’ll remove it.  
It’s to avoid having a too long list of 100% covered files in the output, so we can concentrate on the one not covered.

paveldedik (commented):
@batisteo I found a few issues:

1. I cannot order by clicking on the header of the table (e.g. on Resources > Internal libraries page), is that supposed to work?
2. The `Internal` button on the right side of the page doesn't do anything, and the arrow next to it is also not openning any submenu for me... I mean this one (maybe I don't have compiled some js stuff? shouldn't that compile automatically?):
button
3. I don't see any `CI templates` tab nor `Project templates`, do I need to fill some table with data first? I guess I do, but how? Where do I find the button or something? Or should it be automatic... just asking how it works and what the supposed workflow looks like.

blfpd:
1. I forgot about it. I have to see the how Fomantic work with tables
1. Ooops, I accidently deleted the script dealing with this
1. There are not implemented at all. We would need more work on the models I guess. I might be wrong. Something to develop in other MR pehaps…

paveldedik:
> 3. There are not implemented at all. We would need more work on the models I guess. I might be wrong. Something to develop in other MR pehaps…

OK, I just wanted to make sure I'm not missing something :-)

aexvir:
Yep, CI and Project templates weren't part of this feature. We first need to start tracking them 🙂

paveldedik:
OK, great, it works now, I have a few more things maybe for discussion:

1. I don't like that this is a button, it's for switching between internal and public dependencies, I would do it differently, maybe something like the submenu on service detail page.... button
2. I'm missing some stuff for our internal libraries, like a link to library repository and mainly audit results, when I added support for libraries, I also added support for making checks for libraries. Not sure this was discussed or how it should work now.

aexvir:
1. Service detail? 🤔 I'm not sure about what you mean here, this is all in its own view.

2. Agree with this, but I think we can move that to its own issue and not block this from being merged. Agree?

paveldedik:
I meant this:
Screenshot from 2020-03-13 11-50-47

The problem is that the button looks like it's doing some action (like creating a library). Check the button here ong gitlab which is for merging this pull request, it has dropdown and definitelly looks like action, not switching between tabs...

paveldedik:
And sure, we can move the second point to a new issue.

paveldedik:
Maybe it would help if it were two attached buttons, both with visible text, one active and one inactive, that would solve the problem :-)

aexvir:
Aha, got it. I don't think that the submenu is the way here, as it won't look that good with just two options, but we can definitely improve more on the presentation.

I'd maybe just use a regular dropdown right now, without making it look like a button and we can review it in the future.

@batisteo wdyt?

blfpd:
It indeed valid points!

I’m not sure what is a regular dropdown according to you.

blfpd:
I actually went from dropdown to button because it was basically invisible, and one could easily miss the context.

paveldedik:
What about just two joined buttons, like this:

buttonb

If you disagree, just down vote and I'll merge it as it is now.

blfpd:
That's nice, I like it

paveldedik:
OK, great! The last thing I found is that the bars don't fit in page sometimes, not sure if something is wrong on my laptop or you have the same issue?

Screenshot from 2020-03-16 11-02-27

blfpd:
This might be because the version is `None`.

aexvir:
I think we can remove the versions from here right now, can we just display the last version @batisteo ?

blfpd:
I guess you’re talking about the higher version of the dependency, not in absolute in PyPI or whatsoever.

aexvir:
Yep 🙂

zoo/resources/templates/dependency_overview.html
```diff
@@ -0,0 +1,90 @@
...
+{% endblock %}
+
+{% block content %}
+{% include "resource_header.html" with title=view.title icon=view.icon url=request.path only %}
+<table class="ui sortable very basic table">
+  <thead>
+    <tr>
+      <th scope="col" class="sorted collapsing descending">Usage</th>
+      <th scope="col" class="collapsing">Name</th>
+      <th scope="col" class="">Lastest version</th>
```
aexvir:
```suggestion
      <th scope="col" class="">Latest version</th>
```

aexvir:
@batisteo can you address the last three things that SonarQube is complaining about? And we can merge and release this 🙂

paveldedik:
And please also probably fix this:

Screenshot from 2020-03-17 09-28-01

I guess if there is no version, there is just a square with empty value, which looks weird...

blfpd:
Those rules about `<b>` and `<i>` are outdated and pre-HTML5 _IMHO_.  
https://blog.logrocket.com/youre-using-em-wrong/

aexvir:
We can review them, for now. I'm going to merge this.
Good job @batisteo !

PR #283, opened 2020-02-07
feat: Remove allauth and use only default django auth backend

imunitic:
This MR removes allauth from zoo and uses only the default django auth backend.


aexvir:
LGTM, if I have to say something is that requirements shouldn't have been upgraded, or that should have been done in a different commit. That makes the commit be more meaningful, as right now it removes allauth but it also updates every dependency

paveldedik:
Will non-stuff users be able to log in via the `/admin` page? I think it's currently not possible. If I understand the code, the homepage will have a link to `/admin` but that will by available only for stuff users. Is that intended behavior?

imunitic:
@paveldedik It is not intended. I will fix that. You are right, non-staff members can't login.

imunitic:
I already had accounts login hooked up and the template registration/login.html did not have a form but rather just a button that sent to admin/login. This is now fixed with a login form in the template. Normal users can now login through this form. When you logout you are now redirected to '/'

aexvir (approved):

PR #299, opened 2020-04-24
Frontend build changes

aexvir:


aexvir:
This should lower the amount of frontend processing we are doing

PR #305, opened 2020-06-01
feat: Add notification hook

### LINKED ISSUE #150: Slack notifications for auditing issues
It would be nice to have notifications about `NEW` or `REOPENED` auditing Issues sent to Slack channel specific to Service/team.

In addition our application security team would like to get notifications about `NEW` or `REOPENED` Issues from subset of kinds from all Services to one Slack channel.

cc @janmasarik

blfpd:
- It would be nice to have notifications about NEW or REOPENED auditing Issues sent to Slack channel specific to Service/team.

- In addition our application security team would like to get notifications about NEW or REOPENED Issues from subset of kinds from all Services to one Slack channel.

Fix #150


test/auditing/test_runner.py
```diff
@@ -243,3 +243,18 @@ def test_run_checks_and_save_results__failing_check(repository, fake_path, mocke
...
+
+
+def test_notify_status_change(mocker, issue_factory, service_factory):
+    service = service_factory()
+    issue = issue_factory(repository=service.repository)
+    slack = mocker.Mock()
+    mocker.patch("zoo.auditing.runner.slack", slack)
+    uut.notify_status_change(issue)
+    audit_url = f"services/{issue.repository.owner}/{issue.repository.name}/auditing/"
+    text = "{status} issue {issue} on <{repo.url}|{repo.name}>.".format(
```
aexvir: Hmm, could this be a constant? So we don't have to update the test if we just change the string... wdyt?

aexvir (changes requested):
Nice! I've just left a small comment, other than that it looks good to me 🙂 
Good job!

test/auditing/test_runner.py
blfpd: Sorry I don’t get it actually: even in `text` is static, if the message is change in the runner function, this text has to change. What am I missing?

aexvir (approved):

PR #310, opened 2020-08-24
feat: Add service Tier

blfpd:


zoo/services/models.py
```diff
@@ -26,6 +26,13 @@ class Impact(Enum):
     EMPLOYEES = "employees"
 
 
+class Tier(Enum):
```
aexvir: I'm not a huge fan of having this fully hardcoded, as I can imagine other users for which this tiers might mean nothing 😕 

And also we don't have any description for what every tier means.

I think it might make more sense to add a new `Tier` model, with level and description, and we can enter the tiers through the admin panel for now.

aexvir: and the `tier` in the service model will just be a foreign key to the `tier` table.

aexvir: and we should probably keep it nullable, as not every Zoo user might need to use the tiers

aexvir (changes requested):
As Zoo is an OSS project, and this change shouldn't take that much, I'd much rather keep it more flexible. I don't think we should merge this as it is right now.

zoo/services/models.py
blfpd: I agree, that’s pretty much what I reverted.

aexvir: oh... 😓

aexvir (approved):
Thanks for refactoring this 🙂️
It's much more flexible now, I like it

PR #314, opened 2020-08-28
fix: change version field for dependency usage size from 100 to 200

imunitic:


zoo/analytics/models.py
```diff
@@ -65,7 +65,7 @@ class Meta:
     major_version = models.BigIntegerField(null=True, blank=True)
     minor_version = models.BigIntegerField(null=True, blank=True)
     patch_version = models.BigIntegerField(null=True, blank=True)
-    version = models.CharField(max_length=100, null=True, blank=True)
+    version = models.CharField(max_length=200, null=True, blank=True)
```
aexvir, line 68: I guess when we chose 100 characters it was pretty arbitrary. Do you have some better reason for choosing 200? 😅 
Or will we use this until the next time we need a bigger string?

imunitic: I do think 200 should be enough but I'm not 100% sure. One more thing we can do is to add some truncate logic to [zoo/analytics/tasks/repo_analyzers.py ](https://github.com/kiwicom/the-zoo/blob/master/zoo/analytics/tasks/repo_analyzers.py#L41) ?

Here hit.version length is also pretty random :)

aexvir: Will leave like this for now.

aexvir (approved):

PR #316, opened 2020-09-03
feat(openapi): change source from repo to URL

jarosevcik:
Currently OpenAPI definition is taken from repository, change it to take it from defined URL per environment.


Makefile
```diff
@@ -30,7 +30,7 @@ fake:
 migrate:
 	docker-compose run app python manage.py migrate
 
-makemigrations:
+make migrations:
```
aexvir: Hmm... `make make migrations`? Maybe it would be better to just name `migrations`.

```suggestion
migrations:
```

zoo/repos/views.py
```diff
@@ -35,56 +22,3 @@ def repo_details(request, provider, repo_id):
         raise Http404(f"Project {repo_id} not found")
 
     return JsonResponse(details)
-
-
-def _parse_file(path, base=None):
```
aexvir, line 40: I think we should keep the git filesystem as fallback 🤔 
If the service has an url on the environment we'll load it, if not we'll fallback to check the git repo.

What do you think?

aexvir (changes requested):
Other than that it looks good 🙂

Makefile
jarosevcik: didnt realise that but `make migrations` is working for me 😕

zoo/repos/views.py
jarosevcik: hmm ok,and in case that there is an url on the environment and also repo... we will load the one from environment, right?

aexvir: yep, environment will always have predence.

zoo/services/views.py
```diff
@@ -1,22 +1,26 @@
...
 from django.utils import timezone
 from django.views import generic as generic_views
 from djangoql.exceptions import DjangoQLError
 from djangoql.queryset import apply_search
 
+from .models import Service
 from ..auditing.models import Issue
 from ..checklists.steps import STEPS
 from . import forms, models
+from ..repos.views import openapi_definition_from_repo
```
aexvir: I'd just leave it as `openapi_definition`. It's already clear that it's from a repository once it's on the repos module.

aexvir: ...and also... there is no point anymore on having this function on `views.py`, right? as it won't be directly called via an API call anymore 🤔️

aexvir (changes requested):

zoo/services/views.py
jarosevcik: yop was thinking about moving it to utils.py?

aexvir: sure 🙂️

aexvir (approved):
Thanks! Looks nice 🙂️

PR #320, opened 2020-09-09
feat(checks): add possibility to exclude files or directories 

jarosevcik:
add possibility to exclude files or directories  from security checks


zoo/repos/models.py
```diff
@@ -52,6 +53,13 @@ def languages_from_analytics(self):
             )
         )
 
+    @property
+    def exclude_security_checks(self):
+        try:
+            return Service.objects.get(repository=self).exclude_security_checks
```
aexvir: 🤔 This won't work I'm afraid. The relation between services and repositories is many to one. There can be multiple services that have the same repository.

aexvir: In the future we'll be adding proper monorepo support, and in that case we could be scanning parts of the repository separately.

zoo/services/models.py
```diff
@@ -79,6 +79,12 @@ class Meta:
         default=list,
         help_text="Used for filtering and defining presets of checks",
     )
+    exclude_security_checks = models.CharField(
```
aexvir: We are going to use it for our security checks for now, yes, but this field should be called something more generic, like `exclusions` or something like that.

zoo/repos/models.py
aexvir: For now we either merge the exclusion lists for all the services targeting a repo, or we move the exclusions field to the repository directly.

I think I'm more inclined towards the second option, as otherwise we'd be doing some magic that users won't be able to see.

And we can still have the field in the service form, and updating it in one service will make the change visible on all the other services targeting the same repo.

aexvir: WDYT?

aexvir (changes requested):

zoo/repos/models.py
jarosevcik: in case we plan to add monorepo support it make much more sense for me to keep it on Service model. For now it could be a little confusing(the magic behind it - merging exclusions from multiple service over one repo), but i think it is not any big deal...It could be a bit confusing as well that we will be storing it on repo and displaying it on multiple services - in case that there are multiple people working on same repo, different service what would be the behaviour in case of changing the exclusion list?

tl:dr i prefer keep it on Service but not any strongly... are you ok wit it or you would better go the way you suggested?

aexvir: Yes, it makes more sense on service conceptually. And it will be there for sure once we add monorepo support. But until then we'll have this problem...

As right now we don't really have this issue with our services both things will have virtually the same impact, but the second option seems a bit more natural to me, as the repo is a shared resource between multiple services, and we would have the exclusion list being shared as well.

> in case that there are multiple people working on same repo, different service what would be the behaviour in case of changing the exclusion list?

Well... having services a and b sharing repo; if you modify it on service a, it would also affect service b, and when editing any of them it should be visible on the form, as the exclusion will be a shared resource.

jarosevcik:
Retry DeepCode

zoo/services/forms.py
```diff
@@ -87,6 +95,24 @@ class Meta:
             "tags": TagInput(),
         }
 
+    def __init__(self, *args, **kwargs):
+        if "instance" in kwargs and kwargs["instance"].repository:
+            kwargs["initial"] = {
+                "exclusions": kwargs["instance"].repository.exclusions or ""
+            }
```
aexvir: suggestion: I think this could be simplified like this, so it's a bit easier to read in my opinion. What do you think?

```suggestion
        instance = kwargs.get("instance")
        exclusions = instance.repository.exclusions if instance else ""

        kwargs.update({"initial": {"exclusions": exclusions}})
```

aexvir (commented):
Just a last nitpick

aexvir (changes requested):
Sorry... I wanted to request changes, not approve 🤷

zoo/services/forms.py
aexvir: oops... `repository` isn't a required field in service... so it will need an extra check

aexvir (approved):
Nice! Good job 🙂

PR #332, opened 2020-10-14
Sentry manager methods

blfpd:
1. Register ModelAdmin classes with the decorator
1. Move some code from views to the QuerySet manager of SentryIssue

### DIFF
--- a/zoo/auditing/admin.py
+++ b/zoo/auditing/admin.py
@@ -2,9 +2,8 @@ from django.contrib import admin
 
 from . import models
 
+admin.register(models.Issue)
+
 
 class IssueAdmin(admin.ModelAdmin):
     search_fields = ("kind_key",)
-
-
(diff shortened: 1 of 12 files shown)


zoo/auditing/admin.py
```diff
@@ -2,9 +2,8 @@
 
 from . import models
 
+admin.register(models.Issue)
+
```
aexvir: 🤔 Hmm... shouldn't this also be using the decorator?

```suggestion
@admin.register(models.Issue)
```

aexvir: oh well... GitHub seems to ignore the empty lines on the suggestions... 😅

blfpd: Ups, yep, it seems I missed that one

aexvir (approved):
Thanks a lot! 🙂️

PR #333, opened 2020-10-15
Do not send request for OpenAPI if not URL specified

blfpd:


zoo/services/views.py
```diff
@@ -213,10 +213,12 @@ def dispatch(self, request, *args, **kwargs):
...
             environments = service.environments_dict
-            for _, environment in environments.items():
-                spec = requests.get(environment.open_api_url)
+            urls = [
+                env.open_api_url for env in environments.values() if env.open_api_url
+            ]
+            for url in urls:
+                spec = requests.get(url)
                 specs.append(json.loads(spec.text))
-        specs = list(filter(None, specs))
```
blfpd, line 219: I honestly don’t know why it was there. What was the good reason?

jarosevcik: tbh me neither, it was there before so i just left it there

PR #335, opened 2020-10-15
fix(services): import SentryIssueCategory from zoo models

aexvir:


blfpd:
Closing in favor of #336 because of circular importing

PR #342, opened 2020-11-23
Jgrgicevic/sw 593 envs

gjuro87:


yarn.lock
```diff
@@ -0,0 +1,2 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
aexvir: hmm... we don't need this file

zoo/repos/migrations/0008_auto_20201119_1206.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.2.16 on 2020-11-19 12:06
```
aexvir: let's squash this migration into the previous one

zoo/services/assets/js/service_form.js
```diff
@@ -1,6 +1,14 @@
 import '../style/service_form.less'
+import { gitlabEnvs } from "./gitlab_envs.js"
 
 $(document).ready(function () {
+    //set default env type to zoo
+    $(".ui.segment.environment:visible").find('input[name$="type"]').each(function () {
```
aexvir: the only environments that should be possible to be added/edited via the ui are zoo ones, I'd rather remove the type input completely and set it directly on the backend

what do you think?

zoo/services/migrations/0027_auto_20201123_1128.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.2.16 on 2020-11-23 11:28
```
aexvir: same here, let's squash this 🙂️

zoo/services/assets/js/service_form.js
```diff
@@ -44,15 +53,25 @@ $("button.remove-environment").click(function () {
...
-        var firstSegment = hidden.eq(0);
+        let firstSegment = hidden.eq(0);
         firstSegment.find(":not(.no-reset) input").val("");
         firstSegment.find(".field.name input").prop("required", true);
+        firstSegment.find('input[name$="type"]').val(envTypeZoo);
         firstSegment.show();
     }
 });
+
+$(".load-gitlab-environment").on("click", function () {
```
aexvir: hmm... what about firing this whenever the repository field is updated on the service form instead? what do you think?

zoo/services/assets/js/gitlab_envs.js
```diff
@@ -0,0 +1,110 @@
+const gitlabEnvs = {
```
aexvir, line 1: hmm... this seems really complex, do we need all this? we don't have to reuse the forms for the zoo environments, as the gitlab envs should be read only, so this should be as simple as getting the data from api and displaying some simple divs with the data

aexvir (changes requested):

zoo/repos/tasks.py
```diff
@@ -154,3 +158,24 @@ def get_zoo_file_content(proj: Dict) -> str:
     return provider.get_file_content(
         proj["id"], settings.ZOO_YAML_FILE, settings.ZOO_YAML_DEFAULT_REF
     )
+
+
+def sync_enviroments_from_gitlab(repo: Repository):
+    RepositoryEnvironment.objects.filter(repository_id=repo.id).delete()
```
blfpd: Why not using `objects.get_or_create` instead?

yarn.lock
gjuro87: Removed

zoo/services/migrations/0027_auto_20201123_1128.py
gjuro87: Squashed

zoo/services/assets/js/service_form.js
gjuro87: I have removed button. Now envs are loaded on init and on changed of repo

zoo/services/assets/js/gitlab_envs.js
gjuro87: Ok, i looks like it misunderstood US. Now it only show gitlab envs without using forms for zoo envs

zoo/services/assets/js/service_form.js
gjuro87: I have removed it, yes i doesn't make sense to have it. Also i am thinking to remove type  from Environment model which I introduced because there is new model for gitlab environments called RepositoryEnvironment. And since both type of environments don't mix. I don't see the point to have property type on both models. What do you think?

zoo/repos/tasks.py
gjuro87: Well this task needs to sync envs from gitlab. What if some env was deleted on gitlab. I am not sure that  get_or_create would remove deleted env from db.

zoo/repos/tasks.py
```diff
@@ -208,3 +213,33 @@ def get_zoo_file_content(proj: Dict) -> str:
...
+    ).delete()
+
+    # update gitlab envs on every service
+    services = Service.objects.filter(repository_id=repo.id)
+    for service in services:
+        Environment.objects.filter(
+            service_id=service.id, type=EnviromentType.GITLAB.value
+        ).exclude(name__in=[env.name for env in envs]).delete()
+        for gl_env in gl_envs:
+            _, _ = Environment.objects.update_or_create(
```
blfpd: I don’t think we need the assignment here (`_, _ = `)

zoo/repos/views.py
```diff
@@ -22,3 +24,24 @@ def repo_details(request, provider, repo_id):
...
+        repo = Repository.objects.get(id=project_id, provider=Provider.GITLAB.value)
+    except RepositoryNotFoundError:
+        return JsonResponse({"message": "Wrong project_id"}, safe=False)
+
+    return JsonResponse(
+        [
+            {"name": gl_env.name, "dashboardUrl": gl_env.external_url}
+            for gl_env in repo.repository_environments.all()
+        ],
+        safe=False,
```
blfpd, line 50: I don’t think `self=False` is required, as your passing simple dicts with strings. **(×3)**

gjuro87: It is, i tried to remove it and code won't work

blfpd (approved):

PR #351, opened 2021-01-04
WIP: Jgrgicevic/sentry panel

gjuro87:


zoo/api/types.py
```diff
@@ -1,3 +1,5 @@
+from random import randrange
+
```
blfpd: You just have to remove this and we’re good to go!

frontend/plugins/services/package.json
```diff
@@ -26,6 +26,12 @@
     "@material-ui/lab": "4.0.0-alpha.57",
     "@backstage/core": "^0.4.0",
     "@backstage/theme": "^0.2.2",
+    "@mdi/react": "1.4.0",
+    "@mdi/js": "5.8.55",
+    "numeral": "2.0.6",
+    "@types/numeral": "0.0.29",
+    "recharts": "^1.8.5",
+    "@types/recharts": "^1.8.18",
```
blfpd: Also, could you put `@types/numeral` and `@types/recharts` to devDependencies please?

frontend/plugins/services/src/components/ServiceSentryHistogram.tsx
```diff
@@ -0,0 +1,54 @@
...
+  fontSize: 12,
+  paddingRight: 5,
+  paddingBottom: 1,
+  paddingTop: 1,
+  paddingLeft: 5
+}
+
+type TooltipProps = {
+  active?: boolean;
+  payload?: Array<any>;
```
blfpd, line 17: Maybe `object[];` might fix the DeepCode issue

blfpd: But let’s forget about it. I might change it in another commit.

PR #355, opened 2021-02-02
refactor(pip): Use Poetry for packaging

blfpd:


docker-compose.yml
```diff
@@ -49,7 +49,7 @@ services:
   postgres:
     image: postgres:12-alpine
     volumes:
-      - ./postgres-data:/var/lib/postgresql/data
+      - ./postgres-data:/var/lib/postgresql/data:z
```
jarosevcik, line 52: Whats this :z here?

Dockerfile
```diff
@@ -13,17 +13,22 @@ RUN yarn production
 FROM python:3.8-slim
 
 ENV DJANGO_SETTINGS_MODULE=zoo.base.settings
+ENV POETRY_VERSION=1.1.4
```
jarosevcik, line 16: do we want to pin version?is there any reason why?

blfpd: I’d like to know why not pinning versions.

docker-compose.yml
blfpd: It doesn’t relate to Poetry, it shouldn’t be in this MR. 
It’s useful for Linux with SELinux so that the volume don’t have to be root in the local dev machine. https://stackoverflow.com/a/35222815/5091196

Dockerfile
jarosevcik: checked it up and indeed in docker files its better to pin...

PR #357, opened 2021-02-04
Upgrading CI

blfpd:
- Using pinned `ubuntu-20.04` to avoid warning and future suprises. See https://github.com/actions/virtual-environments/issues/1816
- Updated pre-commit hooks


blfpd:
Retry DeepCode

PR #360, opened 2021-02-05
Jgrgicevic/sw 727 resource list component

gjuro87:


blfpd:
The table search doesn’t pickup the name. I’m not sure how long would it be to fix it.
Do you plan to add this feature in a later commit/MR?

frontend/plugins/resources/src/components/LanguagesPage/LanguagesPage.tsx
```diff
@@ -1,12 +1,91 @@
...
-
-const LanguagesPage = () => (
-  <Page themeId="home">
-    <Content>
-        <div>Welcome to language page!</div>
-    </Content>
-  </Page>
-);
+import {Content, Page, Table, TableColumn, TableFilter} from '@backstage/core';
+import {DummyResponse, Edge, Resource} from "../../../../../packages/zoo-api";
```
blfpd: Did you try with just `from "zoo-api";`? It should work as it is in `package.json`

frontend/plugins/resources/src/components/LanguagesPage/LanguagesPage.tsx
```diff
@@ -1,12 +1,91 @@
...
+import ResourceLink from "../TableComponents/ResourceLink";
+import {Chip, Grid} from "@material-ui/core";
+import ResourceTypeLabel from "../TableComponents/ResourceTypeLabel";
+import Alert from "@material-ui/lab/Alert";
+
+
+const generateTableData: (resources: Array<Resource>) => Array<{}> = (resources) => {
+  const data: Array<{}> = [];
+  for(let i=0; i<resources.length; i++){
```
blfpd: I’d prefer:
```js
for (const resource of resources) {
```

Ref: [for...of in MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of)

…but I guess you could do the same with `.map()`:

```js
return resources.map(resource => ({
	usage: resource.usageCourt,
	…
}))
```

frontend/plugins/resources/src/components/LanguagesPage/LanguagesPage.tsx
```diff
@@ -1,12 +1,91 @@
...
+    if (response.error.message === "[Network] Unauthorized") {
+      return <Grid item>
+        <Alert severity="error">Unauthorized. Please check your Zoo token in the Settings.</Alert>
+      </Grid>;
+    }
+    return <Grid item>
+      <Alert severity="error">{response.error.message}</Alert>
+    </Grid>;
+  }
+  const resources: Resource[] = response.data.resources.edges.map((edge: Edge) => edge.node);
```
blfpd: You could use my handy [`unwrap` function](https://github.com/kiwicom/the-zoo/blob/hangar/frontend/packages/zoo-api/src/client.ts#L32-L40) here:

```js
const resources = unwrap<Resource>(response.data.resources);
```

frontend/plugins/resources/src/components/LibrariesPage/LibrariesPage.tsx
```diff
@@ -1,13 +1,108 @@
...
+    data.push({
+      usage: resources[i].usageCount,
+      name: <ResourceLink name={resources[i].name} id={resources[i].id} />,
+      version: <Chip label={resources[i].version} />,
+      type: <ResourceTypeLabel name={resources[i].type} />,
+      kind: <ResourceKind name={resources[i].name} />,
+    });
+  }
+
+  return data;
```
blfpd: Same comment regarding `return resources.map()` :arrow_up:

frontend/plugins/resources/src/components/TemplatesPage/TemplatesPage.tsx
```diff
@@ -1,13 +1,97 @@
 import React from 'react';
-import { Content, Page } from '@backstage/core';
+import {Content, Page, Table, TableColumn} from '@backstage/core';
+import {DummyResponse, Edge, Resource} from "../../../../../packages/zoo-api";
```
blfpd:
```js
import {…} from "zoo-api";
```

blfpd (changes requested):

frontend/plugins/resources/src/components/LibrariesPage/LibrariesPage.tsx
gjuro87: Fixed, i have used map as you suggested. It is cleaner

frontend/plugins/resources/src/components/LanguagesPage/LanguagesPage.tsx
gjuro87: fixed

frontend/plugins/resources/src/components/LanguagesPage/LanguagesPage.tsx
gjuro87: Used map()

frontend/plugins/resources/src/components/TemplatesPage/TemplatesPage.tsx
gjuro87: This was generated by my IDE, fixed

gjuro87:
> The table search doesn’t pickup the name. I’m not sure how long would it be to fix it.
> Do you plan to add this feature in a later commit/MR?

@batisteo you are right. It doesn't work because i have used components like chip component as a value of table and search is doing compare by value. In case of compoment is object so it doesn’t match it. I will look at source code of backstage table and create our own table that matches our needs.

gjuro87:
> > The table search doesn’t pickup the name. I’m not sure how long would it be to fix it.
> > Do you plan to add this feature in a later commit/MR?
> 
> @batisteo you are right. It doesn't work because i have used components like chip component as a value of table and search is doing compare by value. In case of compoment is object so it doesn’t match it. I will look at source code of backstage table and create our own table that matches our needs.

@batisteo fixed :)

blfpd:
Great!

Could you address the DeepCode issues?

PR #361, opened 2021-02-09
feat: Set graphql endpoint to be local on client

imunitic:
To access the zoo backend graphql endpoint through IAP authentication, and to avoid CORS issue the client will assume that the graphql is on the same domain (/graphql). During deployment we will make sure that /graphql is proxied to the correct zoo backend.

### DIFF
--- a/zoo/base/settings.py
+++ b/zoo/base/settings.py
@@ -21,6 +21,7 @@ version = _get_app_version()
 root = Path(__file__).parents[1]
 env = environ.Env(
     ZOO_DEBUG=(bool, False),
+    ZOO_USE_HANGAR=(bool, False),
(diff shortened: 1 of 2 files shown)


zoo/base/settings.py
```diff
@@ -151,9 +152,12 @@
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
-    "zoo.api.middleware.ApiTokenAuthenticationMiddleware",
 ]
 
+USE_HANGAR = env("ZOO_USE_HANGAR")
+if USE_HANGAR:
```
blfpd: `not`?

blfpd (approved):

PR #362, opened 2021-02-12
feat: add useApiTokenAuth config

imunitic:
This enables us to switch on/off authentication for the
graphql backend as on production we use Googles IAP

### DIFF
--- a/frontend/app-config.yaml
+++ b/frontend/app-config.yaml
@@ -1,6 +1,7 @@
 app:
   title: Hangar App
   baseUrl: http://localhost:3000
+  useApiTokenAuth: false
 
 organization:
   name: Kiwi.com
--- a/frontend/packages/zoo-api/src/client.ts
+++ b/frontend/packages/zoo-api/src/client.ts
@@ -25,8 +25,8 @@ export const theZooClient = new Client({
   url: '/graphql',
   exchanges: [cache, dedupExchange, fetchExchange],
   fetchOptions: () => {
-      let config = useApi(configApiRef);
-      let useApiTokenAuth = config.getBoolean("app.useApiTokenAuth");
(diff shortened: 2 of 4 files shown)


frontend/packages/zoo-api/src/client.ts
```diff
@@ -11,9 +12,30 @@ const cache = cacheExchange({
   },
 });
 
+
+function getToken(): string {
+    const value = localeStorage.getItem("the-zoo.api.token");
```
blfpd: Could you please at least:

1. Move [the constant declaration](https://github.com/kiwicom/the-zoo/blob/hangar/frontend/plugins/services/src/state/AppState.tsx#L9) in this file
2. Rename the constant name
3. Import this constant and rename it [where it’s used](https://github.com/kiwicom/the-zoo/blob/hangar/frontend/plugins/services/src/state/useSettings.ts)

imunitic: this is the exact function as it was before i removed severals commits ago :)

imunitic: done

frontend/app-config.yaml
```diff
@@ -1,6 +1,7 @@
 app:
   title: Hangar App
   baseUrl: http://localhost:3000
+  useApiTokenAuth: false
```
blfpd, line 4: Does it work for you? It seems like we’d need to change the schema of the config somewhere.  
Anyway, it doesn’t pick the config when I’m trying locally :confused:

blfpd (changes requested):

frontend/app-config.yaml
imunitic: I still have to successfully run this locally.

PR #366, opened 2021-02-19
feat: customize graphql for resources frontend

gjuro87:


gjuro87:
Everything is done. There is one issue that i solved in another way. Issue when calling dependency graphql query and requesting id. Result is error: "'Dependency' object has no attribute 'pk'". I couldn't find the cause. I have solve it another way, not requesting id, but needs to be solved. If you could take a look

gjuro87:
Bug fixed, thx for help

frontend/packages/zoo-api/src/queries.ts
```diff
@@ -394,54 +452,24 @@ export interface ActiveIncident extends Node {
   color: string;
 }
 
-export interface Resource extends Node {
+export interface Dependancy extends Node {
```
blfpd: Could you replace Depend**a**ncy by Depend**e**ncy in this MR?

blfpd (changes requested):
Annoying typo. Otherwise good MR! :blush:

blfpd:
For the Pytest check, run `pytest test/api/query/test_node.py --snapshot-update` and review https://github.com/kiwicom/the-zoo/blob/hangar/test/api/query/snapshots/snap_test_node.py

frontend/packages/zoo-api/src/queries.ts
gjuro87: Sorry, I am little bit dyslexic;) Fixed

PR #382, opened 2021-02-24
fix(audit): incorrect audit report redirect params

jarosevcik:


aexvir:
Please let's make sure this is covered by tests 🙂️
Right now it was broken and we didn't have any failing test because of it

blfpd:
And I guess those lines have to be deleted :confused: 
https://github.com/kiwicom/the-zoo/blob/master/.github/workflows/push.yml#L64-L65

test/auditing/test_views.py
```diff
@@ -102,3 +102,32 @@ def test_apply_bulk_patches__has_issues(mocker, client, user, issue_factory):
 
     assert response.status_code == 200
     m_task.delay.assert_called_once_with([issue_1.id, issue_2.id])
+
+
+def test_url_resolution(client, user, service_factory):
```
aexvir: hmm... this seems to be testing django's native `reverse` function, which is probably already tested by django itself 😅️

what I'd test here is that during the execution of `notify_status_change` there is a call to `reverse` with the correct attributes

e.g.
```python
m_reverse.assert_called_once_with('service', service.owner_slug, service.name_slug)
```

aexvir (changes requested):

blfpd (approved):
Looks good

PR #386, opened 2021-03-11
feat(repos): index openapi endpoints

aexvir:
the main feature of this pr is indexing and saving in our db all the openapi endpoints we're obtaining when we're parsing the openapi definitions

while implementing this I was facing numerous issues on the auditing analyzers so I also addressed those, this should clean up some of those annoying sentry errors, plus extending the support for newer / more advanced syntaxes of the things we're analyzing

please review commit by commit... it will be easier imo


jarosevcik (approved):

pyproject.toml
```diff
@@ -131,3 +131,41 @@ tox-docker = ">=2.0.0"
 [build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
+
+
+[too.pylint.master]
```
blfpd: too**l**

blfpd (approved):

PR #390, opened 2021-03-29
fix(repos): keep tmp dir during the checks run

jarosevcik:


zoo/repos/utils.py
```diff
@@ -84,16 +84,24 @@ def _parse_file(path, base=None):
...
-                log.info("repos.views.openapi.git_error", repo=repository, error=err)
-                return JsonResponse(
-                    {"error": "error downloading repository"}, status=500
-                )
+        try:
+            repo_dir = tempfile.TemporaryDirectory()
+        except (FileExistsError, PermissionError) as err:
+            log.info("repos.views.tmpdir", repo=repository, error=err)
+            return JsonResponse(
+                {"error": "error creating temporal directory"}, status=500
```
blfpd:
```diff
- temporal
+ temporary
```

aexvir:
first of all... can you please add tests? 🙏️

zoo/repos/utils.py
```diff
@@ -84,16 +84,24 @@ def _parse_file(path, base=None):
...
-            except (MissingFilesError, RepositoryNotFoundError) as err:
-                log.info("repos.views.openapi.git_error", repo=repository, error=err)
-                return JsonResponse(
-                    {"error": "error downloading repository"}, status=500
-                )
+        try:
+            repo_dir = tempfile.TemporaryDirectory()
+        except (FileExistsError, PermissionError) as err:
+            log.info("repos.views.tmpdir", repo=repository, error=err)
+            return JsonResponse(
```
aexvir: not related to your changes, but this isn't correct
the function should only return a list, the responses will be handled on the views

so... change both exception handlings to return empty list

aexvir (approved):

PR #392, opened 2021-04-07
feat: zoo search

gjuro87:


zoo/api/query.py
```diff
@@ -142,3 +151,53 @@ def resolve_all_dependency_usages(self, info, **kwargs):
         return types.DependencyUsageConnection(
             page_info=page_info, edges=edges, total_count=total
         )
+
+    def resolve_all_search_results(self, info, **kwargs):
+        query_param = kwargs["search_query"] if "search_query" in kwargs else ""
+
+        meili_client = meilisearch.Client(
```
jarosevcik: what about initialising only one global client?

zoo/meilisearch/views.py
```diff
@@ -1,48 +1,96 @@
...
+            results = meili_client.get_index(index["uid"]).search(
+                query=search_query, opt_params={"offset": offset, "limit": limit}
             )
-            objects_to_return.update(objects_for_index)
-        return objects_to_return
+            objects_to_return[f"total_{index['name'].lower()}"] = results["nbHits"]
+
+            if index_type == index["uid"]:
+                new_offset = results["offset"]
+                total_hits = results["nbHits"]
```
aexvir: this should have a default... otherwise it won't be defined if the condition isn't met, so it will break on line 51

zoo/api/query.py
```diff
@@ -142,3 +153,80 @@ def resolve_all_dependency_usages(self, info, **kwargs):
...
+            total_services_count=total_count[IndexType.Service.value],
+            total_analytics_count=total_count[IndexType.Dependency.value],
+        )
+
+    @staticmethod
+    def _get_all_for_index(query_param, search_index, total_count):
+        offset = 0
+        limit = 100
+        search_results = []
+        while total_count != len(search_results):
```
aexvir: I don't like this... there isn't any protection against endless loops 😕 
what happens if for some reason the query doesn't return any results?

personally I'd prefer to have all the pagination logic either done inside this function or resolved when this function is called, to not rely on endless loops

example
```python
def _get_all_for_index(query_param, index_uid):
    limit = 100
    results = []
    total = _get_hit_count(query_param, index_uid)
    pages = math.ceil(total / limit)

    for page in range(pages):
        response = meili_client.get_index(index_uid).search(
            query=query_param,
            opt_params={"offset": page * limit, "limit": limit}
        )
        results.append(
            Query._objects_from_result(response["hits"], search_index, search_results)
    return results
```

zoo/api/query.py
```diff
@@ -142,3 +153,80 @@ def resolve_all_dependency_usages(self, info, **kwargs):
...
+            results = meili_client.get_index(search_index["uid"]).search(
+                query=query_param, opt_params={"offset": offset, "limit": limit}
+            )
+
+            search_results = Query._objects_from_result(
```
aexvir: this should be appending, right?

zoo/api/query.py
```diff
@@ -142,3 +153,80 @@ def resolve_all_dependency_usages(self, info, **kwargs):
...
+            model = apps.get_model(index["uid"], index["name"])
+            for i, result in enumerate(search_results):
+                object = model.objects.get(pk=result["id"])
+                if index["name"] == IndexType.Dependency.name:
+                    dependency = types.Dependency.from_db(object)
+                    node = types.SearchResult(id=i, dependency=dependency)
+                if index["name"] == IndexType.Service.name:
+                    service = types.Service.from_db(object)
+                    node = types.SearchResult(id=i, service=service)
+                result_objects.append(node)
```
aexvir, line 227: if `result_objects` is `None`, as it's specified on the default, this will fail 🤔 

overall I don't like the approach of calling this function with a list, it will cause unwanted side effects, like modifying the original array while also returning it, is this necessary?

just make it build and return a list without passing any pointer to it and handle merging lists wherever this is called

zoo/api/query.py
aexvir: oh, I see now... it's modifying the `search_results` variable inside the function 😢 

this is a no-no for me

zoo/meilisearch/views.py
```diff
@@ -1,48 +1,96 @@
-import meilisearch
 import structlog
 from django.apps import apps
-from django.conf import settings
 from django.views.generic import TemplateView
 
+from .meilli_client import meili_client
```
aexvir: typo on `meilli_client`

zoo/meilisearch/views.py
```diff
@@ -1,48 +1,96 @@
...
+    def convert_meili_to_pages(self, total_hits, offset, limit=meili_limit):
+        if total_hits < limit:
+            return {
+                "total_pages": 1,
+                "current_page": 1,
+                "next_page": None,
+                "previous_page": None,
+            }
+
+        total_pages = int(total_hits / limit)
```
aexvir: I'd rather use `math.ceil` to make obvious that we want the upper bound of the division result

aexvir (changes requested):
there are a couple things to iron out before merging

zoo/api/query.py
```diff
@@ -142,3 +153,80 @@ def resolve_all_dependency_usages(self, info, **kwargs):
         return types.DependencyUsageConnection(
             page_info=page_info, edges=edges, total_count=total
         )
+
+    def resolve_all_search_results(self, info, **kwargs):
+        paginator = Paginator(**kwargs)
+        query_param = kwargs["search_query"] if "search_query" in kwargs else ""
```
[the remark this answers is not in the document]
gjuro87: fixed

zoo/api/query.py
```diff
@@ -142,3 +153,80 @@ def resolve_all_dependency_usages(self, info, **kwargs):
...
+        search_type = (
+            kwargs["search_type"]
+            if "search_type" in kwargs
+            else IndexType.Service.value
+        )
```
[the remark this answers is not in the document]
gjuro87: fixed

zoo/api/query.py
gjuro87: Yes, this part all together is clumsy, i have made changes as you suggested. It looks better.

zoo/meilisearch/views.py
gjuro87: fixed

zoo/meilisearch/views.py
gjuro87: fixed

PR #393, opened 2021-04-15
jaro/Introduce Arbitrary Links for Service

jarosevcik:


test/repos/test_zoo_yml.py
```diff
@@ -65,10 +76,17 @@ def test_generate(generate_services_with_environments):
...
   dashboard_url: https://dashboardurl
-  health_check_url: https://healthcheckurl
+  health_check_url: null
   service_urls:
   - https://serviceurl1
   - https://serviceurl2
+arbitrary_links:
+- name: Datadog
+  url: https://dashboard.datadog.com
+  emoji: ':simple_smile:'
```
jarosevcik: Either we have to live with fact that emoji will be dumped with `''` around it or we have to store it without `::`

aexvir: hmm... how are we rendering the emojis on the frontend? 🤔 
I don't see any frontend change to display this data 😄️ 
so... I'd defer this decision until we decide how we'll display them

zoo/repos/zoo_yml.py
```diff
@@ -53,6 +53,17 @@
                             type: string
                     health_check_url:
                         type: ["string", "null"]
+        arbitrary_links:
```
aexvir: I'd call it just `links` tbh 😅

aexvir (changes requested):
overall it looks ok, but just having the model won't be enough I'd say... we need to show the links on the ui and also add them to the api response

zoo/base/assets/style/project_detail.less
```diff
@@ -59,6 +59,10 @@
     max-width: 80%;
   }
 
+  .sub.header.links{
```
aexvir: this isn't needed anymore, is it?

zoo/repos/zoo_yml.py
```diff
@@ -100,4 +112,12 @@ def generate(service: Service) -> str:
         }
         result["environments"].append(environ)
 
+    for link in service.links.all():
+        link = {
+            "name": link.name,
+            "url": link.url,
+            "icon": link.icon,
+        }
+        result["links"].append(link)
```
aexvir: nitpick: i personally like it more like this 😄️ 
```suggestion 
        result["links"].append(
	          {
	              "name": link.name,
	              "url": link.url,
	              "icon": link.icon,
	          }
        )
```

zoo/services/forms.py
```diff
@@ -60,6 +60,14 @@ class Meta:
         }
 
 
+class LinkForm(forms.ModelForm):
+    class Meta:
+        model = models.Link
+        fields = ["name", "url", "icon"]
+
+    labels = {"name": "Name", "url": "URL", "icon": "Emoji"}
```
aexvir: it's not emoji, so it might not need the labels at all in the end
```suggestion
    labels = {"name": "Name", "url": "URL", "icon": "Icon"}
```

zoo/services/templates/services/service_detail.html
```diff
@@ -290,5 +310,6 @@ <h4 class="ui {{ color }} marginless header">
   </script>
 {% endverbatim %}
 {{ block.super }}
+<script src="js/accordion.js"></script>
```
aexvir: hmm... what's this?

aexvir (changes requested):
I didn't expect the service form to support this feature, we agreed on having it on the yml definitions only.

But it's a welcome addition nonetheless 😄️ 

I've left a couple comments here and there, overall looks good to me

aexvir (approved):

PR #405, opened 2021-06-17
Entity model 

jarosevcik:


zoo/entities/migrations/0001_initial.py
```diff
@@ -0,0 +1,150 @@
+# Generated by Django 2.2.19 on 2021-07-01 11:30
```
aexvir: once more, migration name 😄 
and I think we should only have a migration on this commit, as we're just adding the models here, so let's squash them into one

zoo/base/apps.py
```diff
@@ -24,7 +24,7 @@ def ready(self):
 
         celery_app.add_periodic_task(timedelta(hours=1), repos_tasks.sync_repos)
         celery_app.add_periodic_task(timedelta(hours=1), repos_tasks.schedule_pulls)
-        celery_app.add_periodic_task(timedelta(days=1), repos_tasks.sync_zoo_file)
+        # celery_app.add_periodic_task(timedelta(days=1), repos_tasks.sync_zoo_file)
```
aexvir: delete please, we can always get it back from git, no need to have it commented out

aexvir: but... don't we want to have a task for syncing entities? 🤔

test/entities/test_yaml_to_objects.py
```diff
@@ -0,0 +1,262 @@
...
+    assert component_entity.type == "library"
+    assert component_entity.links.all().count() == 2
+    assert component_entity.group
+    assert component_entity.service is None
+    assert component_entity.library is not None
+
+
+def test_create_multiple_components(mocker, repository_factory):
+    multiple_components = """
+- apiVersion: v1alpha1
```
aexvir: this shouldn't be a list
in yaml, having multiple areas separated by `---` means multiple documents inside the same yaml file, which should be enough for what we're trying to do here, and it will also be more consistent with [backstage](https://backstage.io/docs/features/software-catalog/descriptor-format#overall-shape-of-an-entity)

so...
```yaml
apiVersion: v1alpha1
kind_: component
metadata:
  name: base
# morestuff
---
apiVersion: v1alpha1
kind_: component
metadata:
  name: service
# morestuff
```

zoo/entities/yaml_definitions/component_library.yaml
```diff
@@ -0,0 +1,97 @@
+---
+additionalProperties: false
+properties:
+  apiVersion:
+    type: string
+  kind_:
```
aexvir: why `kind_`? 🤔

zoo/entities/yaml_definitions/component_library.yaml
```diff
@@ -0,0 +1,97 @@
...
+          type: object
+        type: object
+      lifecycle:
+        enum:
+          - beta
+          - production
+          - deprecated
+          - discontinued
+        type: string
+      type_:
```
aexvir: and `type_`?

zoo/repos/entities_yaml.py
```diff
@@ -0,0 +1,149 @@
...
+    "base_component": str(Path(__file__).parents[1])
+    + "/entities/yaml_definitions/component_base.yaml",
+    "service": str(Path(__file__).parents[1])
+    + "/entities/yaml_definitions/component_service.yaml",
+    "library": str(Path(__file__).parents[1])
+    + "/entities/yaml_definitions/component_library.yaml",
+}
+
+
+def select_schema(entity):
```
aexvir: `get_schema` seems more suitable

zoo/entities/yaml_definitions/component_library.yaml
aexvir: is it because of conflicting with `type` on jsonschema? but there shouldn't be any problem with that... am I missing something

and if that's it then there should be better ways to fix it than forcing our users to add random underscores to some fields 😅

test/entities/test_yaml_to_objects.py
aexvir: or do we have any other reason for needing a list here? from what I've seen everywhere is read as `entity[0]` all the time

zoo/entities/builder.py
```diff
@@ -0,0 +1,107 @@
...
+class EntityBuilder:
+    def create_entity(self, data, repository):
+        if data["kind_"] != "component":
+            return NotImplemented
+        if data["spec"]["type_"] == "service":
+            self._build_service(data, repository)
+        elif data["spec"]["type_"] == "library":
+            self._build_library(data, repository)
+        else:
+            self._build_base_component(data, repository)
```
aexvir, line 17: do we have this use case? of storing just a base component? 🤔

aexvir: I only see `create_entity` called with entities, for the base we already have `_build_base_component`

test/entities/test_yaml_to_objects.py
```diff
@@ -0,0 +1,262 @@
...
+    repository_dict = {"id": repository.remote_id, "provider": repository.provider}
+    mocker.patch(
+        "zoo.repos.tasks.get_entity_file_content", return_value=multiple_components
+    )
+    uut.update_project_from_entity_file(proj=repository_dict)
+    component_entity = Entity.objects.first()
+    assert Entity.objects.all().count() == 2
+    assert Service.objects.all().count() == 1
+
+    # assert component_entity.kind == "component"
```
aexvir: same here, no comments on the final mr 🙏🏻

zoo/repos/entities_yaml.py
```diff
@@ -0,0 +1,149 @@
...
+        }
+        component_base_doc["spec"]["environments"].append(environment_document)
+
+    return component_base_doc
+
+
+def generate(repository: Repository) -> str:
+    results = []
+    for entity in repository.entities.all():
+        if entity.kind != "component":
```
aexvir, line 135: let's store the kind in an enum 🙏🏻

aexvir (changes requested):

aexvir:
also... I'm totally up for having multiple commits to make this easier to review as there are two big chunks of work, first changing the models then updating the api, but right now those 5 commits we have don't really seem meaningful, so please squash/rename/split what's needed to have it clear

aexvir:
and don't forget about the `README.md` 😅 
it's still mentioning the old schema

zoo/services/constants.py
```diff
@@ -1,7 +1,7 @@
 from enum import Enum
 
 
-class Status(Enum):
+class Lifecycle(Enum):
```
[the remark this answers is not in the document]
jarosevcik: Shall i then move the lifecycle to schema metadata?Also thinking about the initial application of this changes, we need to keep the lifecycle on service and library, then generate yaml files, import it as entity where we can store it on Entity model and then we can remove the fields from service and library... or can you think the different approach?

test/entities/test_yaml_to_objects.py
jarosevcik: 🤦 and i was confused why the hell i have to use `entity[0]` ... of course you are correct and it should not be list...

zoo/entities/builder.py
jarosevcik: IIRC we were talking about letting people store also other types than service and library but we wont be storing any `spec` info about them... for now i added the `database` only, but we can definitely think about more of this types. So that was my use case for base component

imunitic: I don't like this if chain here. I would personally replace it with a registry pattern and create builders for service and library as separate classes. Then pass the registry to the Entity builder to use it. This way adding new component builders will be simple to add and simple to understand.

imunitic: One personal rule I have is "Private methods usually indicate that those should be extracted as public methods of a separate class"

zoo/repos/entities_yaml.py
```diff
@@ -0,0 +1,151 @@
...
+    "service": str(Path(__file__).parents[1])
+    + "/entities/yaml_definitions/component_service.yaml",
+    "library": str(Path(__file__).parents[1])
+    + "/entities/yaml_definitions/component_library.yaml",
+}
+
+
+def get_schema(entity):
+    if entity["kind"] == "component":
+        if entity["spec"]["type"].lower() == "service":
```
imunitic, line 28: Can you eliminate these 2 ifs with just a single one?
In pseudo-code something like:
```
if contains(SCHEMAS, entity["spec"]["type"].lower()) then
    return SCHEMAS[entity["spec"]["type"].lower()]
endif
```

jarosevcik: technically yes, logically it was ment to separate `kind` and `type`. Now we have only `component` kind but could have more in future...

zoo/entities/builder.py
jarosevcik: it mentioned in TODO as i didnt want to spend more time on this, could be refactored in future when adding new types/kinds

imunitic: 👍

zoo/repos/entities_yaml.py
imunitic: I was referencing the 2 type ifs being replaced. You leave the kind if

